PR #3, opened 2021-08-31
Crossref Events Maintenance

jdddog:
This PR fixes a few issues with Crossref Events:
* Uses ProcessPoolExecutor rather than ThreadPoolExecutor for transforming files. Processes are faster than threads for transforming files.
* Run on remote_queue by default.
*  Replace @ symbol in keys, due to @ symbol left in keys by DataCite between the 15 and 22 March 2019.


aroelo (changes requested):
Hey Jamie,
Looks good to me, but the new transform_batch function is not covered by the unit tests.
I wonder if those lines are actually covered, but not picked up somehow because of the processpoolexecutor?

jdddog:
> Hey Jamie,
> Looks good to me, but the new transform_batch function is not covered by the unit tests.
> I wonder if those lines are actually covered, but not picked up somehow because of the processpoolexecutor?

Thanks Aniek, ah maybe, I'll take a look

jdddog:
> Hey Jamie,
> Looks good to me, but the new transform_batch function is not covered by the unit tests.
> I wonder if those lines are actually covered, but not picked up somehow because of the processpoolexecutor?

Yeah that seems to be the case, I've changed that last test a little bit so now that it is covered.

aroelo (approved):

PR #6, opened 2021-08-31
ORCID maintenance

jdddog:
This PR adds a few fixes for the ORCID telescope:
* Run on `remote_queue`.
* Use ProcessPoolExecutor rather than ThreadPoolExecutor for transforming files, as processes are generally faster for IO bound tasks.
* Transform files in batches of 500. The VM kept on running out of memory when queuing all of the file transforms (there are a lot with ORCID) into the ProcessPoolExecutor, so now they are processed in batches to prevent this.


aroelo (approved):
Looks good, thanks Jamie.

PR #26, opened 2021-09-02
Add elastic index retention policy

tuanchien:
Add index keep info to the elastic import workflow for deleting stale indices

### DIFF
--- a/academic_observatory_workflows/dags/elastic_import_workflow.py
+++ b/academic_observatory_workflows/dags/elastic_import_workflow.py
@@ -19,15 +19,16 @@
 
 import json
 import os
-from typing import List
+from typing import Callable, Dict, List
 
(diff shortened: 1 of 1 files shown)


academic_observatory_workflows/dags/elastic_import_workflow.py
```diff
@@ -95,5 +104,6 @@ def load_elastic_mappings_ao(path: str, table_prefix: str, simple_prefixes: List
         elastic_mappings_func=config.elastic_mappings_func,
         kibana_spaces=config.kibana_spaces,
         kibana_time_fields=config.kibana_time_fields,
+        index_keep_info=index_keep_info,
```
jdddog, line 107: Shouldn't this be `index_keep_info=config.index_keep_info`?

jdddog (approved):

PR #29, opened 2021-09-03
Remove usage of custom AirflowVariable class

aroelo:
The custom AirflowVariable class has been removed in the observatory-platform, this PR updates usage of that custom class to Airflow's standard Variable class instead.


jdddog (approved):
Thanks Aniek, it looks good to me!

PR #35, opened 2021-09-13
Bugfix deployment

jdddog:
This PR addresses the following:
* Crossref Metadata now needs the API key to make a head request, so I've added it to the head request.
* Made a standardised `make_release_date` function and use this in Onix Workflow. It returns a pendulum.DateTime rather than a pendulum.Date.
* <s>Fixed a bug in requirements.sh, i.e. removed "RUN" before the echo command </s>: merged here instead https://github.com/The-Academic-Observatory/academic-observatory-workflows/pull/37

The tests will fail until this PR is merged: https://github.com/The-Academic-Observatory/observatory-platform/pull/494

### DIFF
--- a/academic_observatory_workflows/workflows/crossref_metadata_telescope.py
+++ b/academic_observatory_workflows/workflows/crossref_metadata_telescope.py
@@ -64,6 +64,12 @@ class CrossrefMetadataRelease(SnapshotRelease):
         self.url = CrossrefMetadataTelescope.TELESCOPE_URL.format(year=release_date.year, month=release_date.month)
 
     @property
+    def api_key(self):
+        """ Return API token """
+        connection = BaseHook.get_connection(AirflowConns.CROSSREF)
+        return connection.password
+
+    @property
     def download_path(self) -> str:
         """Get the path to the downloaded file.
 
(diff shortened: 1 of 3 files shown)


academic_observatory_workflows/workflows/crossref_metadata_telescope.py
```diff
@@ -276,7 +280,10 @@ def check_release_exists(self, **kwargs):
         url = CrossrefMetadataTelescope.TELESCOPE_URL.format(year=execution_date.year, month=execution_date.month)
         logging.info(f"Checking if available release exists for {execution_date.year}-{execution_date.month}")
 
-        response = retry_session().head(url)
+        # Get API key: it is required to check the head now
```
rhosking, line 283: could this use the api_key() method?

jdddog: Ah, not in this case because the release hasn't been created at this point

rhosking: ahh yeah, makes sense. Opened up the whole file (not just the diff) and went through it again. Thanks

rhosking (approved):
looks good

PR #39, opened 2021-09-24
Inf 40/create unpaywall data feed telescope

tuanchien:
Unpaywall data feed telescope uses a stream telescope template.
- Fetches daily changefiles
- Start date should be set to the snapshot date you want to base dataset on and be after (2021,7,2).
- Daily update runs will fetch the daily changefile from 2 days prior to the execution_date.  This is so that the first changefile fetched is 1 day prior to the snapshot date (to guarantee data integrity).  In effect, this means the dataset on a given execution_date is 2 days behind the latest changefile.

### DIFF
--- a/academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
+++ b/academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
@@ -36,6 +36,7 @@ from observatory.platform.utils.url_utils import (
     get_http_response_json,
     get_observatory_http_header,
 )
+from observatory.platform.utils.workflow_utils import is_first_dag_run
 from observatory.platform.workflows.stream_telescope import (
     StreamRelease,
     StreamTelescope,
--- a/academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
+++ b/academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
@@ -99,11 +99,11 @@ class UnpaywallDataFeedRelease(StreamRelease):
         """Download the release."""
         # May need to refactor depending on how snapshot looks.
         if self.first_release:
-            self.download_snapshot_()
+            self._download_snapshot()
         else:
-            self.download_data_feed_()
+            self._download_data_feed()
 
-    def download_snapshot_(self):
+    def _download_snapshot(self):
         """Download the most recent Unpaywall snapshot on or before the start date."""
 
         download_file(url=self.snapshot_url, headers=self.http_header, prefix_dir=self.download_folder)
@@ -138,7 +138,7 @@ class UnpaywallDataFeedRelease(StreamRelease):
 
         return (None, None)
 
-    def download_data_feed_(self):
+    def _download_data_feed(self):
         """Download data feed update (diff) that can be applied to the base snapshot.  Can only handle a single download."""
 
         url, filename = self.get_diff_release(
@@ -194,7 +194,7 @@ class UnpaywallDataFeedTelescope(StreamTelescope):
         :param airflow_vars: list of airflow variable keys, for each variable it is checked if it exists in airflow
         """
 
-        self.validate_schedule_interval_(start_date=start_date, schedule_interval=schedule_interval)
+        self._validate_schedule_interval(start_date=start_date, schedule_interval=schedule_interval)
 
         if airflow_vars is None:
             airflow_vars = [
@@ -233,7 +233,7 @@ class UnpaywallDataFeedTelescope(StreamTelescope):
         self.add_task_chain([self.bq_delete_old, self.bq_append_new, self.cleanup], trigger_rule="none_failed")
 
     @staticmethod
-    def schedule_days_apart_(
+    def _schedule_days_apart(
         *, start_date: pendulum.DateTime, schedule_interval: Union[str, timedelta, relativedelta]
     ) -> Generator:
         """Calculate the scheduled days apart.
@@ -256,7 +256,7 @@ class UnpaywallDataFeedTelescope(StreamTelescope):
             a = b
             yield diff
 
-    def validate_schedule_interval_(self, *, start_date: pendulum.DateTime, schedule_interval: str):
+    def _validate_schedule_interval(self, *, start_date: pendulum.DateTime, schedule_interval: str):
         """Check that the schedule interval gives us 1 or 7 day differences.
         Throws exception on failure.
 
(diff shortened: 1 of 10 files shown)


academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def download(self):
+        """Download the release."""
+        # May need to refactor depending on how snapshot looks.
+        if self.first_release:
+            self.download_snapshot_()
+        else:
+            self.download_data_feed_()
+
+    def download_snapshot_(self):
```
jdddog, line 106: Private methods in Python should begin with an underscore, not end with an underscore, e.g. _download_snapshot.

academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
```diff
@@ -0,0 +1,326 @@
...
+            # Apply diffs from 2 days ago.  This is so we start applying diffs 1 day before the snapshot date to
+            # guarantee no gaps with the snapshot.
+            target_date = (start_date - pendulum.Duration(days=2)).date()
+
+            if release_date == target_date:
+                return release["url"], release["filename"]
+
+        return (None, None)
+
+    def download_data_feed_(self):
```
jdddog, line 141: Private methods in Python should begin with an underscore, not end with an underscore, e.g. _download_snapshot.

academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
```diff
@@ -0,0 +1,326 @@
...
+        self.add_task(self.upload_downloaded)
+        self.add_task(self.extract)
+        self.add_task(self.transform)
+        self.add_task(self.upload_transformed)
+        self.add_task(self.bq_load_partition)
+
+        self.add_task_chain([self.bq_delete_old, self.bq_append_new, self.cleanup], trigger_rule="none_failed")
+
+    @staticmethod
+    def schedule_days_apart_(
```
jdddog, line 236: Private methods in Python should begin with an underscore, not end with an underscore, e.g. _download_snapshot.

academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
```diff
@@ -0,0 +1,326 @@
...
+        a = start_date
+        it = croniter(schedule_interval, start_date)
+
+        while True:
+            b = it.next(datetime)
+            diff = (b - a).days
+            a = b
+            yield diff
+
+    def validate_schedule_interval_(self, *, start_date: pendulum.DateTime, schedule_interval: str):
```
jdddog, line 259: Private methods in Python should begin with an underscore, not end with an underscore, e.g. _download_snapshot.

academic_observatory_workflows/workflows/unpaywall_data_feed_telescope.py
```diff
@@ -0,0 +1,326 @@
...
+        start_date, first_release = self.get_release_info_(**kwargs)
+        release = UnpaywallDataFeedRelease(
+            dag_id=self.dag_id,
+            start_date=start_date,
+            end_date=start_date,
+            first_release=first_release,
+        )
+        return release
+
+    def get_release_info_(self, **kwargs) -> Tuple[pendulum.DateTime, bool]:
```
jdddog, line 317: Private methods in Python should begin with an underscore, not end with an underscore, e.g. _download_snapshot.

docs/telescopes/unpaywall_data_feed.md
```diff
@@ -0,0 +1,62 @@
...
+2. Get all changefiles starting with latest timestamp just before the snapshot date.
+3. Apply the changefiles to the snapshot in date order from oldest to newest.
+
+The Data Feed service requires an API key in order to access the changefiles.  See the [data feed page](https://unpaywall.org/products/data-feed)
+for more information on obtaining a key.
+
+On first run, this telescope tries to pull the snapshot on the telescope's start date.  Subsequent scheduled runs will download the daily changefile from __TWO DAYS PRIOR__ to the scheduled execution date to update the dataset.  You should set the scheduled_interval to "@daily" or some other equivalent interval which that results in scheduled runs on a daily basis. The telescope attempts to catch up on missed scheduled runs in case of interruption.
+
+The reason that the changefile applied is from two days prior to each scheduled executin date is so that we can guarantee data integrity after applying the snapshot.
+
```
jdddog, line 29: It would be good to put an example the required Airflow Connection in the documentation, i.e. the key name and the connection URI.

jdddog (changes requested):
It looks good to me Tuan, just a few small comments.

jdddog (approved):
Thanks Tuan, looks good to me!

aroelo:
@tuanchien Did you test whether the BigQuery functionality works with catchup=True?
Witch catchup set to True, there could be multiple dag runs on the same day. 
The partitions that are created in the bq_load_partition task are ingestion time partitions, meaning that the partition date (and table id) is based on the ingestion date, which could possibly be the same for multiple Dag Runs.
I'm not sure if the data is then properly handled or if it is e.g. overwritten?

tuanchien:
@aroelo I did not test that bit. You're right, I think there is an issue.  However, according to https://cloud.google.com/bigquery/docs/managing-partitioned-table-data#write-to-partition we can target explicit partition dates even with ingestion based partitioning.

I haven't looked closely into the details of the stream telescope, and its bq functions, but perhaps, if we change
https://github.com/The-Academic-Observatory/observatory-platform/blob/64b62cba83110fea60e91506dff4a83ba9931ba9/observatory-platform/observatory/platform/utils/workflow_utils.py#L440

from:
```
partition_table_id = create_date_table_id(partition_table_id, pendulum.today(), partition_type)
```

to
```
partition_table_id = create_date_table_id(partition_table_id, release.start_date, partition_type)
```
that might do it (```end_date``` is another option).

aroelo:
> https://github.com/The-Academic-Observatory/observatory-platform/blob/64b62cba83110fea60e91506dff4a83ba9931ba9/observatory-platform/observatory/platform/utils/workflow_utils.py#L440

Yeah I think changing it to the release.end_date would be a good alternative, I've made that change in the PR that is already open and related to the StreamTelescope (https://github.com/The-Academic-Observatory/observatory-platform/pull/496/commits/fced3e713de3e41eac22190e5214c2af21c252f3)

PR #43, opened 2021-09-27
Inf 121/crossref metadata update

jdddog:
This PR addresses the following:
* More robust Crossref Metadata transform step:
  * Can transform JSON input files whether pretty printed or not.
  * Refactored transform functions to use Python rather than mawk as it is easier to understand what the transformation does and it is faster. 11 hrs 16 mins for mawk on latest dataset and 3 hrs 41 mins with Python. I suspect it would be faster if the JSON input files were not pretty printed.
  * Added explicit tests for transform_file and transform_item.
* Latest Crossref Metadata schema.
* Fix bug in Fundref get_release_info, which was passing a string to bigquery_sharded_table_id rather than a DateTime.

### DIFF
--- a/academic_observatory_workflows/database/schema/crossref_metadata_2021-07-07.json
+++ b/academic_observatory_workflows/database/schema/crossref_metadata_2021-07-07.json
@@ -1991,7 +1991,7 @@
   {
     "mode": "NULLABLE",
     "name": "score",
-    "type": "NUMERIC",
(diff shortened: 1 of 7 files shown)


rhosking (approved):
I like the move into Python, I kickstarted the bash transform script a few years back as a quick fix. It has served us well, but this looks much easier to test (not to mention quicker to run)

PR #52, opened 2021-10-26
INF-60: Open Access Website Workflow

jdddog:
This PR adds a workflow that generates the data for the COKI Open Access website, copies the data to the websites static folder, builds the website and then deploys it with Cloudflare.

### DIFF
--- a/academic_observatory_workflows/clearbit.py
+++ b/academic_observatory_workflows/clearbit.py
@@ -1,0 +1,38 @@
+# Copyright 2021 Curtin University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Author: James Diprose
+
+import shutil
+
+import requests
+
+
+def clearbit_download_logo(*, company_url: str, file_path: str, size: int = 24, fmt: str = "jpg") -> bool:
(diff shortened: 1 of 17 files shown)


aroelo (changes requested):
Hi @jdddog,

I've just added a small comment about the variables/connections.
It might also be good to add a readthedocs file with some info on the website_folder variable and how to create the cloudflare api connection.

Also, before we split the repository for the different workflow projects we would add new variables/connections in the AirflowVars & AirflowConns classes in airflow_utils.py.
What is the recommended approach now that we have split the repositories?
Should we only include those variables/connections that are used with templates or utils inside the observatory-platform, so remove ORCID, OAPEN IRUS UK, etc.
Or do we still add any new variables/connections here, like the ones used in this workflow?

And just curious myself, what is the website_folder and what do those build & deploy scripts look like, is that in a different repository somewhere?

jdddog:
> Hi @jdddog,
> 
> I've just added a small comment about the variables/connections. It might also be good to add a readthedocs file with some info on the website_folder variable and how to create the cloudflare api connection.

Thanks Aniek, will do.

> 
> Also, before we split the repository for the different workflow projects we would add new variables/connections in the AirflowVars & AirflowConns classes in airflow_utils.py. What is the recommended approach now that we have split the repositories? Should we only include those variables/connections that are used with templates or utils inside the observatory-platform, so remove ORCID, OAPEN IRUS UK, etc. Or do we still add any new variables/connections here, like the ones used in this workflow?

It would be good to put the workflow specific variables and connections into each workflow as class variables. We should remove the others from the observatory platform.

> 
> And just curious myself, what is the website_folder and what do those build & deploy scripts look like, is that in a different repository somewhere?

The website folder is the folder where the static website is cloned into. The build script builds the website and the deploy script deploys it to Cloudflare. Yeah I'll be making a project soon, with the web application for the open access dashboard which will contain these scripts.

jdddog:
I've made those changes Aniek. I've remove the need for the website folder being specified as an Airflow variable.

aroelo:
> I've made those changes Aniek. I've removed the need for the website folder being specified as an Airflow variable.

Thanks @jdddog, looks good to me!

aroelo (approved):

PR #54, opened 2021-10-29
Remove Unpaywall history

jdddog:


rhosking (approved):
looks good. I don't think is used downstream, so shouldn't have any side effects

PR #55, opened 2021-10-29
Inf 164/port wos telescope template

tuanchien:
Web of Science telescope template port.

For the Airflow test conducted, the details that are normally fetched from the API was substituted for hard coded values instead.


jdddog (approved):
Thanks Tuan, it looks good!

PR #60, opened 2021-11-09
INF-65: MAG updates

jdddog:
This is an update to enable MAG to be ingested after the changes to how it is shared on Azure. Now all of the MAG releases are deposited on a single container and the Google Storage Transfer service enables you to specify a location in the bucket to transfer the files to, which has enabled some simplification.

Since MAG will stop at the end of the year, there isn't much point in porting it to use the workflow template to simplify it further or to write tests for it.


rhosking (approved):
looks good!

PR #64, opened 2021-11-18
Inf 183 - Move to ROR

rhosking:
Changes the primary identifier for Institutions from GRID to ROR. Grid IDs are still what is provided by MAG, so we still use the external_ids.GRID.preferred_id in the RoR dataset to perform the match.

The pull request also removes the home_url field and corresponding upstream logic as it was no longer being used, and not being updated. Future work will address a better approach to this.


jdddog (approved):
Thanks Richard, it looks good to me.

PR #66, opened 2021-11-23
INF-287: Update to work with Next.js web application

jdddog:


aroelo (approved):
Hey @jdddog,

It all looks good to me!

I'm having a little bit of trouble understanding some of the steps in the transform task myself. 
Perhaps you were already planning for this, but it might be good to write some documentation on what this workflow achieves in general lines.
This could also include some details on e.g. the construction of the Cloudflare Airflow connection.

Some questions for me that popped up:
- Why is the index saved in 3 different formats as JSON, JSONL and Parquet?
- Do all the files in the website_build_folder end up somewhere on Cloudflare as part of the build/deploy steps? So I could see what the files look like there.
- What are the entities, autocomplete and stats?

We could just chat on slack as well if that is easier :-)

aroelo (approved):
Nice work @jdddog, sounds like a good idea to build from a github action instead.
While I was working on the wikipedia descriptions I noticed there is no clean up task for this workflow.
Should we add one so the local files are deleted from the VM? There is already a cleanup method for the Workflow parent class that could be called.

PR #67, opened 2021-11-25
INF 290 schemas for doi workflow

rhosking:
added schemas and descriptions to derived tables in the DOI workflow. These schemas will not currently be used in table creation, that will be a different piece of work


jdddog (changes requested):
Thanks Richard, they look good. Would you be able to add descriptions for each of the fields?

jdddog:
Also, would you be able to use the schema so that the descriptions are loaded into the BigQuery tables?

rhosking:
closing this pull request, as I have pre-merged this in to the documentation branch (because I wanted to render the schemas on readthedocs)

PR #68, opened 2021-11-29
INF-208 Updated DOI workflow page in readthedocs

rhosking:
WIP to create a documentation page detailing the DOI workflow


docs/workflows/doi.md
```diff
@@ -1,4 +1,120 @@
-# DOI
+# DOI Workflow
 
-Add Workflow Description here
+The primary purpose of the DOI workflow is to link together open datasets of higher education through the use of persistent identifiers, and to then produce useful summary statistics across a variety of common aggregations
 
+## Input datasets
+
+Each a number of specialised research publication datasets are collected. These include Crossref Metadata, Crossref Funder Registry, Crossref Events, Microsoft Academic Graph, Unpaywall, the Research Organization Registry, Open Citations and Geonames. The table below illustrates each of these datesets and the PIDs that are exposed within them.
```
kathrynnapier, line 7: Suggested change to 'A number of '.

docs/workflows/doi.md
```diff
@@ -1,4 +1,120 @@
...
+
+``` eval_rst
+.. image:: ../static/diagrams/pid_workflow.png
+   :width: 600
+```
+
+Once the publications table has been created, the publications are grouped by entity type and publication year. For instance publications are grouped by institution and publication year. The last step involves creating aggregate timeseries statistics based on the yearly groups of publications.
+
```
kathrynnapier, line 55: Should the label 'papers' in this figure be 'publications'?

docs/workflows/doi.md
```diff
@@ -1,4 +1,120 @@
...
+``` eval_rst
+.. image:: ../static/diagrams/aggregate_publications.png
+   :width: 650
+```
+
+### Intermediate Datasets
+
+Digging into the mechanics of the workflow, there are a number of intermediate tables that are created. Much of the time, these are not relevant as the information is better organised in the final output tables. However, for specific analysis they may be useful so are included here for completeness 
+
```
kathrynnapier, line 64: Can you explain why the intermediate tables are created?

docs/workflows/doi.md
```diff
@@ -1,4 +1,120 @@
...
++----------------+----------------------------------------------------+-----+-----------------------------------+
+| MAG            | `observatory_intermediate.magYYYYMMDD`             | DOI | create_mag.sql.jinja2             |
++----------------+----------------------------------------------------+-----+-----------------------------------+
+| Open Citations | `observatory_intermediate.open_citationsYYYYMMDD`  | DOI | create_open_citations.sql.jinja2  |
++----------------+----------------------------------------------------+-----+-----------------------------------+
+| ORCID          | `observatory_intermediate.orcidYYYYMMDD`           | DOI | create_orcid.sql.jinja2           |
++----------------+----------------------------------------------------+-----+-----------------------------------+
+| Unpaywall      | `observatory_intermediate.unpaywallYYYYMMDD`       | DOI | create_unpaywall.sql.jinja2       |
++----------------+----------------------------------------------------+-----+-----------------------------------+
+
```
kathrynnapier, line 83: Is it possible to link directly to the  location of the scripts?

docs/workflows/doi.md
```diff
@@ -1,4 +1,120 @@
...
+```
+
+### Output datasets
+
+The final set of output dataset are as followed. Aside from the DOI table, which is an integrated set of all the input datasets linked by DOI, they each share a very similar, or identical, schema. The difference is around the entity being aggregated against.
```
kathrynnapier, line 88: 'The final set of output dataset are as followed' change to 'The final set of output datasets are as follows'.

Schema would be very useful and is requested- if the schemas are identical, use a common table and add the fields that are specific to each table.

docs/workflows/doi.md
```diff
@@ -1,4 +1,120 @@
...
++-------------------+-----------------------------------------------------+-----------+
+| Open Citations    | `open_citations.open_citationsYYYYMMDD`             | DOI       |
++-------------------+-----------------------------------------------------+-----------+
+| ORCID             | `orcid.orcid`                                       | ORCID/DOI |
++-------------------+-----------------------------------------------------+-----------+
+| Unpaywall         | `our_research.unpaywall`                            | DOI       |
++-------------------+-----------------------------------------------------+-----------+
+| iso3166           | `ios.iso3166_countries_and_regions`                 | Country   |
++-------------------+-----------------------------------------------------+-----------+
+```
```
aroelo, line 31: Some minor comments:
- Is the Geonames table the same as the iso3166 table? Or how is this table derived from Geonames?
- Grid is listed, but ROR isn't, I understand we're in the process of transitioning between them, but it is a bit confusing now as ROR is listed in the diagram below, while GRID isn't.
- ROR and/or GRID should also have 'Country' as a PID I think? If I understand correctly that this is how an item is linked to a specific country from iso3166 dataset.

aroelo (commented):
Hi @rhosking, the documentation is very helpful so far and I think it gives a good global overview on the workflow.
I'm having a bit of a closer look at the workflow and for me these two things are still unclear/missing in the documentation:

- There is a dashboard dataset with tables and views, is this for the data studio dashboards? Maybe these should be included in the documentation as well, or will these be removed/deprecated soon?
- There are also the elasticsearch tables in the data_export dataset. I think it would be useful to document these, why they exist and how they are different from the 'final' datasets.

docs/workflows/doi.md
rhosking: yeah, that is probably correct. The diagram was reused from someone Jamie created for the website. Can update it if that other diagram wants to be changed too?

docs/workflows/doi.md
rhosking: Thanks for the feedback! I have added a section on the dashboard and data_export datasets. Now that RoR is merged, I have updated it all to only reference RoR. I also added country as a PID as I think you are right, and that is consistent with how I was using it for other datasets

docs/workflows/doi.md
kathrynnapier: Yup, please do

jdddog (approved):
Thanks Richard

PR #72, opened 2021-12-17
Test fixes after bq quota checks feature

tuanchien:
- fix make_observatory_api import path
- mock out bq daily quota check


jdddog (approved):
Thanks Tuan

PR #77, opened 2022-01-20
Fix crossref metadata invalid timestamp

aroelo:
Some entries have an invalid date_time field, resulting in a BigQuery loading error (I've made the date_time field a STRING here to display the faulty entry).
image

This PR adds a Try/Except check to verify the value for a 'date_time' field. This field is of the 'TIMESTAMP type in BigQuery.
If the value is not a valid datetime, the 'date_time' field will be set to null.

Other date related fields like 'timestamp' (INTEGER) and 'date_parts' (ARRAY of INTEGER) in the same STRUCT remain unchanged, regardless whether 'date_time' is valid or not.

So the above result would now be:
image

I've changed the fixture and updated the unit tests to test for an invalid date_time, however this does not seem to be picked up by codecov. I suspect that is because it is run from the ProcessPoolExecutor.


jdddog (approved):
Thanks for fixing it Aniek, it looks good to me.

jdddog (approved):
Thanks for the fix Aniek!

PR #79, opened 2022-02-01
INF-328: Ignore unknown values on BigQuery load

jdddog:
Enable easier maintenance of snapshot telescopes by setting ignore_unknown_values to True.

This means that the production system will not break due to schema additions to datasets. When we want to use new fields from datasets we will need to explicitly update the schema and the next version of a given dataset will have the updated schema (or a current one if we overwrite it).


aroelo (approved):
Thanks @jdddog

PR #80, opened 2022-02-07
OpenAlex telescope

aroelo:
Adding the OpenAlex telescope.

Things to note:
* OpenAlex has two fields that are objects (RECORDs) with dynamic field names, this makes it impossible to create a schema beforehand and upload the data straight into BigQuery. The fields are abstract_inverted_index (in Works entity only) and international (in Concepts and Institutes entities).
As a workaround, these fields are transformed into a RECORD of two arrays of the same length. The first array contains all the original field names and the second array the corresponding values.

* The point above may not require a workaround if BigQuery allows for JSON data type fields. Currently, this feature is in preview and only for files that are imported from csv (https://cloud.google.com/bigquery/docs/reference/standard-sql/json-data)

* The ‘updated_date’ field of an entry does not always match the date from the folder name that entry sits in.

* The google cloud transfer service has the option to use a manifest file. This would be fitting to use to specify which files needs to be transferred, as OpenAlex makes a manifest file available. Unfortunately, there are permission issues when using the manifest file, this seems to be a Google Cloud issue and I have reported this here https://issuetracker.google.com/issues/216057461.
For now we can use the 'includePrefixes' parameter instead to specify which files need to be transferred, however there is a limit on the number of prefixes you can pass on, which is max 1000. So far the total number of files in the openalex repository is below this number (648 files at this moment).

I have done a full run testing both the first and second release on an external VM.
The results from the first run are in the Utrecht GCP. 
There may have been some small changes since those runs and I have not done a complete test run since, because it is relatively expensive and time consuming to do so.
However, I don't expect there to be any issues when merging this in.
 
* I have also edited the generate_schema_csv.py file, because the OpenAlex schema files are in a separate subdirectory. I could restructure them like e.g. the MAG schema files, but thought it is clearer when the files have their own subdir.

### DIFF
--- a/academic_observatory_workflows/workflows/openalex_telescope.py
+++ b/academic_observatory_workflows/workflows/openalex_telescope.py
@@ -178,6 +178,9 @@ class OpenAlexRelease(StreamRelease):
 
     def download_transferred(self):
         """Download the updated entities from the Google Cloud download bucket to a local directory using gsutil.
+        Gsutil is used instead of the standard Google Cloud Python library, because this is faster at downloading files.
+        It supports multi-threading with the '-m' flag and can open multiple simultaneous connections to GCS.
+        In future the 'gcloud storage' command might be used instead which is even faster, but still in preview.
 
         :return: None.
         """
@@ -193,7 +196,7 @@ class OpenAlexRelease(StreamRelease):
 
         logging.info(f"Downloading transferred files from Google Cloud bucket: {self.download_bucket}")
         log_path = os.path.join(self.download_folder, "cp.log")
-        # Download all records from bucket
+        # Download all records from bucket using Gsutil
         args = [
             "gsutil",
             "-m",
--- a/academic_observatory_workflows/workflows/openalex_telescope.py
+++ b/academic_observatory_workflows/workflows/openalex_telescope.py
@@ -1,0 +1,445 @@
+# Copyright 2022 Curtin University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Author: Aniek Roelofs
+
+from __future__ import annotations
+
+import gzip
+import json
+import logging
+import os
+import subprocess
+from concurrent.futures import ProcessPoolExecutor, as_completed
+from subprocess import Popen
+from typing import List, Tuple
+
+import boto3
+import jsonlines
+import pendulum
+import pytz
+from airflow.exceptions import AirflowException, AirflowSkipException
+from airflow.hooks.base import BaseHook
+from airflow.models.variable import Variable
+from google.cloud import storage
+from observatory.platform.utils.airflow_utils import AirflowVars
+from observatory.platform.utils.gc_utils import aws_to_google_cloud_storage_transfer
+from observatory.platform.utils.proc_utils import wait_for_process
+from observatory.platform.workflows.stream_telescope import (
+    StreamRelease,
+    StreamTelescope,
+)
+
+from academic_observatory_workflows.config import schema_folder as default_schema_folder
+
+
+class OpenAlexRelease(StreamRelease):
+    def __init__(
+        self,
+        dag_id: str,
+        start_date: pendulum.DateTime,
+        end_date: pendulum.DateTime,
+        first_release: bool,
+        max_threads: int,
+        max_processes: int,
+    ):
+        """Construct a OpenAlexRelease instance
+
+        :param dag_id: the id of the DAG.
+        :param start_date: the start_date of the release.
+        :param end_date: the end_date of the release.
+        :param first_release: whether this is the first release that is processed for this DAG
+        :param max_threads: Max threads used for parallel downloading
+        :param max_processes: max processes for transforming files.
+        """
+        super().__init__(
+            dag_id, start_date, end_date, first_release, download_files_regex=".*.gz", transform_files_regex=".*.gz"
+        )
+        self.max_threads = max_threads
+        self.max_processes = max_processes
+
+    @property
+    def transfer_manifest_path_download(self) -> str:
+        """Get the path to the file with ids of updated entities.
+
+        :return: the file path.
+        """
+        return os.path.join(self.download_folder, "transfer_manifest_download.csv")
+
+    @property
+    def transfer_manifest_path_transform(self) -> str:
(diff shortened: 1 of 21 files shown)


docs/telescopes/openalex.md
```diff
@@ -0,0 +1,177 @@
...
+> * Authors are people who create works
+> * Venues are journals and repositories that host works
+> * Institutions are universities and other orgs that are affiliated with works (via authors)
+> * Concepts tag Works with a topic
+>
+> Together, these make a huge web (or more technically, heterogeneous directed graph) of hundreds of millions of entities and over a billion connections between them all.
+
+See https://docs.openalex.org/ for more information.
+
+This telescope transfers OpenAlex data from an AWS s3 bucket and loads it into multiple tables in BigQuery, with one
```
jdddog, line 17:
```suggestion
This telescope transfers OpenAlex data from an AWS S3 bucket and loads it into multiple tables in BigQuery, with one 
```

docs/telescopes/openalex.md
```diff
@@ -0,0 +1,177 @@
...
+> * Institutions are universities and other orgs that are affiliated with works (via authors)
+> * Concepts tag Works with a topic
+>
+> Together, these make a huge web (or more technically, heterogeneous directed graph) of hundreds of millions of entities and over a billion connections between them all.
+
+See https://docs.openalex.org/ for more information.
+
+This telescope transfers OpenAlex data from an AWS s3 bucket and loads it into multiple tables in BigQuery, with one 
+table for each entity (Works, Authors, Venues, Institutions, Concepts).  
+The first run will process all the files that are available in the s3 bucket.
```
jdddog, line 19:
```suggestion
The first run will process all the files that are available in the S3 bucket. 
```

docs/telescopes/openalex.md
```diff
@@ -0,0 +1,177 @@
...
+This means that the files for these entities are directly transferred to the transform bucket.
+
+The other entities do require some transformation and those files are transferred to the download bucket.
+After transforming the data the resulting files are then uploaded to the transform bucket.
+
+The transformation that is required has to do with two fields that have nested fields with dynamic field names.
+These make it impossible to create a schema beforehand and upload the data straight into BigQuery. 
+The two mentioned fields are 'abstract_inverted_index' (present in Work entity only) and 'international' (present in 
+Concept and Institute entities).
+As a workaround, these fields are transformed into a RECORD of two arrays of the same length.
```
jdddog, line 33:
```suggestion

As a workaround, these fields are transformed into a RECORD of two arrays of the same length. 
```

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -0,0 +1,465 @@
...
+        download_transfer = {"manifest": self.transfer_manifest_path_download, "bucket": self.download_bucket}
+        transform_transfer = {"manifest": self.transfer_manifest_path_transform, "bucket": self.transform_bucket}
+
+        total_count = 0
+        for transfer in [download_transfer, transform_transfer]:
+            success = False
+            prefixes = []
+            with open(transfer["manifest"], "r") as f:
+                for line in f:
+                    prefixes.append(line[1:-2])
```
jdddog, line 151: It would be good to have a comment with an example of what the manifest lines look like and what is being cut from each line.

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -0,0 +1,465 @@
...
+            "gsutil",
+            "-m",
+            "-q",
+            "cp",
+            "-L",
+            log_path,
+            "-r",
+            f"gs://{self.download_bucket}/telescopes/{self.dag_id}/{self.release_id}/*",
+            self.download_folder,
+        ]
```
jdddog, line 207: Is the reasoning behind using gsutil rather than using the Google Cloud Python library to download the files because you can sync only the modified files with gsutil?

academic_observatory_workflows/workflows/tests/test_openalex_telescope.py
```diff
@@ -0,0 +1,560 @@
...
+                        )
+                        if run == self.first_run:
+                            self.assertEqual(dag.default_args["start_date"], start_date)
+                            self.assertEqual(pendulum.today("UTC") - timedelta(days=1), end_date)
+                            self.assertTrue(first_release)
+                        else:
+                            self.assertEqual(release.end_date + timedelta(days=1), start_date)
+                            self.assertEqual(pendulum.today("UTC") - timedelta(days=1), end_date)
+                            self.assertFalse(first_release)
+
```
jdddog, line 188: I personally think that utilising loops and if statements for large blocks of unit tests makes them difficult to follow, and that it is preferable for the tests to be more verbose (with the tradeoff of duplication of unit testing code), so that it is easier and quicker to understand how a particular part of the system is tested. The unit tests are partly a recipe for another programmer to learn how the system works.

That is just my opinion though, so don't feel like you have to change it.

jdddog (changes requested):
Thanks Aniek, it looks good, just a few comments. 

It looks like there might be a dependency missing for the docs, now that pandas is removed.

academic_observatory_workflows/workflows/openalex_telescope.py
aroelo: The main reason is that it is faster to download files than using the Google Cloud Python library.
I remember doing some timings/runs with both when initially developing the ORCID telescope, which has a similar download step. I didn't write down the exact timings etc, but gsutil was more efficient.
This post doesn't have any timings either, but confirms the same thought https://stackoverflow.com/questions/44975080/is-there-any-advantage-to-using-gsutil-or-the-google-cloud-storage-api-in-produc

I'm not sure whether the Python library has the features of resuming a download using a log file and/or only copying files that are specified in a text file.

While trying to find some timing numbers I saw that there is also 'gcloud storage' in preview, which presumably has faster download/upload times, maybe something to look into.
https://cloud.google.com/blog/products/storage-data-transfer/new-gcloud-storage-enables-super-fast-data-transfers

academic_observatory_workflows/workflows/openalex_telescope.py
aroelo: I've added an example of a line in the manifest file in the method above, and changed the code so it is clear that only quotes and a newline character are stripped.

`prefixes.append(line.strip("\n").strip('"'))`

academic_observatory_workflows/workflows/tests/test_openalex_telescope.py
aroelo: Mmm yeah, thank you for the feedback. I think you're right and will reformat a bit.

academic_observatory_workflows/workflows/openalex_telescope.py
jdddog: I see, ok that makes sense.

jdddog: Perhaps just put a comment (what you wrote above) there about why gsutil is being used, for future reference.

jdddog (changes requested):

jdddog (approved):
Thanks Aniek!

PR #83, opened 2022-02-11
Inf 333/publish oa web data

jdddog:
Publish the Open Access Web data to a Google Cloud Storage bucket so that it can be pulled down by a Github Action for building the website.

### DIFF
--- a/academic_observatory_workflows/workflows/oa_web_workflow.py
+++ b/academic_observatory_workflows/workflows/oa_web_workflow.py
@@ -43,20 +43,21 @@ import requests
 from airflow.exceptions import AirflowException
 from airflow.models.variable import Variable
 from airflow.sensors.external_task import ExternalTaskSensor
-from observatory.platform.utils.airflow_utils import AirflowVars
+from pyarrow import json as pa_json
+
+from academic_observatory_workflows.clearbit import clearbit_download_logo
+from observatory.platform.utils.airflow_utils import AirflowVars, get_airflow_connection_password
 from observatory.platform.utils.config_utils import module_file_path
 from observatory.platform.utils.file_utils import load_jsonl
 from observatory.platform.utils.gc_utils import (
     bigquery_sharded_table_id,
     download_blobs_from_cloud_storage,
     select_table_shard_dates,
+    upload_file_to_cloud_storage,
 )
 from observatory.platform.utils.workflow_utils import make_release_date
 from observatory.platform.workflows.snapshot_telescope import SnapshotRelease
 from observatory.platform.workflows.workflow import Workflow
-from pyarrow import json as pa_json
-
-from academic_observatory_workflows.clearbit import clearbit_download_logo
 
 # The minimum number of outputs before including an entity in the analysis
 INCLUSION_THRESHOLD = 1000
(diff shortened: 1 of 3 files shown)


docs/workflows/oa_web.md
```diff
@@ -1,6 +1,6 @@
 # Open Access Website Workflow
-A workflow that generates the data for the COKI Open Access Website (still in progress), copies the data to the website
-static folder, builds the website and then deploys it to Cloudflare.
+A workflow that generates the data for the COKI Open Access Website and uploads the data to a Google Cloud Storage
+bucket, which is used to build the Next.js web application.
```
aroelo, line 4:
```suggestion
Each time the workflow runs successfully a repository dispatch event is created, this event triggers the Github actions responsible for building the web application
```
Or something along those lines, I think it might be useful to add a sentence to clarify that the github actions are triggered from this workflow. and that those actions build the web application.

academic_observatory_workflows/workflows/oa_web_workflow.py
```diff
@@ -316,6 +318,32 @@ def clean_ror_id(ror_id: str):
     return ror_id.replace("https://ror.org/", "")
 
 
+def trigger_repository_dispatch(*, token: str, event_type: str, client_payload: Dict = None):
```
aroelo, line 321: Could we get rid of the client_payload param? It does not seem to be in use.

docs/workflows/oa_web.md
```diff
@@ -16,31 +16,56 @@ static folder, builds the website and then deploys it to Cloudflare.
...
 
-```yaml
-cloudflare_api_token: http://:<cloudflare_api_token>@
-```
+### oa_web_data_bucket
+The name of the Google Cloud Storage bucket where the data will be uploaded.
 
-## External references
-* [Cloudflare Creating API tokens](https://developers.cloudflare.com/api/tokens/create)
+When you create the bucket, enable object versioning to keep previous releases.
```
aroelo, line 71: It would be good to add info on the Github token Airflow Conn as well, does this require any specific permissions?

aroelo (changes requested):
Hey @jdddog, it looks good, just a few small questions/comments.

Is the version that is included in the blob name also used for the github actions?
I don't see it passed on anywhere, but I can't find the github actions, maybe it retrieves it when pulling the file from the bucket?

Is the object versioning used for this bucket? I'm not too familiar with it, but I think this is relevant when uploading blobs with the same name?
Aren't the blob names always unique, because they are in a different transform folder for each release?

What is the reasoning for using a separate bucket instead of our standard download/transform, is this related to the object versioning and/or retention time of the files?

academic_observatory_workflows/workflows/oa_web_workflow.py
jdddog: Yep sure

docs/workflows/oa_web.md
jdddog: Thanks, I've added that

jdddog:
> Is the version that is included in the blob name also used for the github actions? I don't see it passed on anywhere, but I can't find the github actions, maybe it retrieves it when pulling the file from the bucket?

Yeah the Github Action pulls from a specific dataset version: https://github.com/The-Academic-Observatory/coki-oa-web/blob/develop/.github/workflows/build-on-data-update.yml#L68-L74. So that when we make breaking changes to the data, the web application won't break.

> 
> Is the object versioning used for this bucket? I'm not too familiar with it, but I think this is relevant when uploading blobs with the same name? Aren't the blob names always unique, because they are in a different transform folder for each release?
> 

The blob names in this case are always the same, i.e. "v1/latest.zip" file, which means that the Github Action can always pull the file down from the same path without any complex logic.

> What is the reasoning for using a separate bucket instead of our standard download/transform, is this related to the object versioning and/or retention time of the files?

The reason is so that we can use object versioning to keep a history of the data files (which has to be applied to the whole bucket) so that we can roll back if necessary. It also means that the logic for the Github Action can be simpler, since the file name is always known.

docs/workflows/oa_web.md
jdddog: Yeah good point, I've added some instructions and will share the current token with you as it is already created.

aroelo (approved):
Thanks @jdddog, looks good.

PR #84, opened 2022-02-11
Inf 326/dataset release creation

tuanchien:
Depends on [INF-325/dataset_release_utils](https://github.com/The-Academic-Observatory/observatory-platform/tree/INF-325/dataset_release_utils)

- Adds basic dataset release record creation in the api for new successful data ingestion runs in snapshot, stream telescopes, and the doi workflow.

### DIFF
--- a/academic_observatory_workflows/workflows/orcid_telescope.py
+++ b/academic_observatory_workflows/workflows/orcid_telescope.py
@@ -259,6 +259,7 @@ class OrcidTelescope(StreamTelescope):
         airflow_vars: List = None,
         airflow_conns: List = None,
         max_processes: int = min(32, os.cpu_count() + 4),
+        workflow_id: int = None,
     ):
         """Construct an OrcidTelescope instance.
 
@@ -275,6 +276,7 @@ class OrcidTelescope(StreamTelescope):
         :param airflow_vars: list of airflow variable keys, for each variable it is checked if it exists in airflow
         :param airflow_conns: list of airflow connection keys, for each connection it is checked if it exists in airflow
         :param max_processes: Max processes used for parallel transforming.
+        :param workflow_id: api workflow id.
         """
         if table_descriptions is None:
             table_descriptions = {
@@ -307,6 +309,7 @@ class OrcidTelescope(StreamTelescope):
             airflow_vars=airflow_vars,
             airflow_conns=airflow_conns,
             batch_load=batch_load,
+            workflow_id=workflow_id,
             load_bigquery_table_kwargs={"ignore_unknown_values": True},
         )
         self.max_processes = max_processes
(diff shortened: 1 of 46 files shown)


academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -498,7 +501,7 @@ def write_modified_record_blobs(
                     last_modified_date = pendulum.parse(elements[3])
 
                     # skip records that are too new, not included in this release
-                    if last_modified_date > end_date:
+                    if last_modified_date >= end_date:
```
jdddog, line 504: @aroelo will this change affect anything?

jdddog (commented):
Thanks Tuan, looks good, just a question for @aroelo about the ORCID telescope.

academic_observatory_workflows/workflows/orcid_telescope.py
aroelo: Depending on whether the start & end date of releases has changed it should be okay. Is this the case with the API changes? I know we have talked about making some changes there but not sure about the final decision/changes @tuanchien 

It used to be like this, as an example for 2 DAG runs:
start date: 01-01-2022
end date: 08-01-2022

start date: 08-01-2022
end date: 15-01-2022

If there is an ORCID record with a modified date of 08-01-2022, this would be included in the second run. So the `last_modified_date >= end_date` statement is fine.

tuanchien: In this PR, `get_release_info` has been changed to this:
```
    def get_release_info(self, **kwargs) -> Tuple[pendulum.DateTime, pendulum.DateTime, bool]:
        """Return release information with the start and end date. [start date, end date) Includes start date, excludes end date.
        :param kwargs: The context passed from the PythonOperator.
        :return: None.
        """

        # Check if first release or not
        first_release = is_first_release(self.workflow_id)
        if first_release:
            # When first release, set start date to the start of the DAG
            # Can we get away with using data_interval_start instead?
            start_date = pendulum.instance(kwargs["dag"].default_args["start_date"]).start_of("day")
        else:
            datasets = get_datasets(workflow_id=self.workflow_id)
            releases = get_dataset_releases(dataset_id=datasets[0].id)
            latest_release = get_latest_dataset_release(releases=releases)
            start_date = pendulum.instance(latest_release.end_date).start_of("day")

        end_date = pendulum.instance(kwargs["data_interval_end"])
        logging.info(f"Start date: {start_date}, end date: {end_date}, first release: {first_release}")

        return start_date, end_date, first_release
```
Apart from very first one,  the next start date is just the end date of the previous release. The release period has a half open interval [start, end) that excludes the end.

Since it's exclusive of the end, ```last_modified_date >= end_date``` means ```last_modified_date``` is in the next release period. I assume that's what your check was doing.

I do have a question about:
```
if first_release:
   start_date = pendulum.instance(kwargs["dag"].default_args["start_date"]).start_of("day")
```
How important is it to keep that behaviour for the first release?

jdddog: We might be able to replace it with `data_interval_start` I think we should just double check that it works as we expect it to.

jdddog (approved):
Thanks Tuan, looks good to me.

PR #85, opened 2022-02-14
Fix ORCID telescope gcloud error

aroelo:
Set CLOUDSK_PYTHON env to fix gcloud error, see https://issuetracker.google.com/issues/217589135


jdddog (approved):
Thanks for the fix Aniek!

PR #86, opened 2022-02-21
Small fixes OA web workflow wiki descriptions

aroelo:
* Updating the expected descriptions after the shortening + removing of text between brackets
* Adding code to handle institutions that do not have an url (alhough I think every institution has an url so far?)
* Add testcase in unit test for an institution without an url
* Always use at least 1 sentence for the wiki description

### DIFF
--- a/academic_observatory_workflows/fixtures/oa_web_workflow/country_wiki_descriptions.json
+++ b/academic_observatory_workflows/fixtures/oa_web_workflow/country_wiki_descriptions.json
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6e7e3c6e5c13b8d4929e90a557f0a9b5b33a6dc39e0bb3953f3742be93603c48
-size 56007
+oid sha256:4630cefc287c842dd70225f760751b263f44da97bdeba611b46dc8294ebf466c
+size 4894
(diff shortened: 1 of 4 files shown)


jdddog (changes requested):
Thanks Aniek, it looks good.

One thing that I noticed is that in the function save_index should be modified so that the description isn't saved, because the description isn't used in any of the tables:
```python
def save_index(self, category: str, df_index_table: pd.DataFrame):
    """Save the index table.
    :param category: the category, i.e. country or institution.
    :param df_index_table: the index table Pandas Dataframe.
    :return: None.
    """
    # Save subset
    base_path = os.path.join(self.build_path, "data")
    os.makedirs(base_path, exist_ok=True)
    df_index_table = df_index_table.drop(
        [
            "description",
            "year",
            "date",
            "institution_types",
            "identifiers",
            "collaborators",
            "subjects",
            "other_platform_locations",
            "timeseries",
        ],
        axis=1,
        errors="ignore",
```

aroelo:
> Thanks Aniek, it looks good.
> 
> One thing that I noticed is that in the function `save_index` should be modified so that the description isn't saved, because the description isn't used in any of the tables. The description is only required in the individual country and institution json files.
> 
> ```python
> def save_index(self, category: str, df_index_table: pd.DataFrame):
>     """Save the index table.
>     :param category: the category, i.e. country or institution.
>     :param df_index_table: the index table Pandas Dataframe.
>     :return: None.
>     """
>     # Save subset
>     base_path = os.path.join(self.build_path, "data")
>     os.makedirs(base_path, exist_ok=True)
>     df_index_table = df_index_table.drop(
>         [
>             "description",
>             "year",
>             "date",
>             "institution_types",
>             "identifiers",
>             "collaborators",
>             "subjects",
>             "other_platform_locations",
>             "timeseries",
>         ],
>         axis=1,
>         errors="ignore",
> ```

Thanks @jdddog, I've just changed this.

jdddog (approved):

PR #87, opened 2022-02-21
Replace MAG with OpenAlex

aroelo:
From @cameronneylon 
> Modifications to create_doi.sql.jinja2 and create_openalex.sql.jinja2 to switch from MAG to openalex
> 
> Queries have been validated but not fully tested against the version of the OpenAlex Works table I have access to. The full implications of the create_doi changes have not been tested at this point.


cameronneylon:
Some observations - I believe that the current pull request represents a non-breaking change for downstream queries (I'm not absolutely sure on how to test this effectively tho) in that all it does in practice is shift the ROR gathering across to OpenAlex (previously from MAG) and add the openalex elements in their own struct. 

We would quite possibly lose coverage, but that's a consequence of the shift in dataset. An alternative is to combine RORs from both MAG and OA but at least in theory there is no reason why OpenAlex isn't a superset of MAG.

tuanchien:
create_doi changes:
- RoR ids pulled from OpenAlex Work table to replace mag grid ids
- openalex Work table embedded in doi table
- Field of study tags switched over to OpenAlex Concept tags

create_openalex changes:
- Creates a new intermediate table called 'openalex' using the Work table
- doi have the "https://doi.org/" prefix trimmed off
- Implements a heuristic for handling the "multiple doi assignment" problem.

Heuristic:
- In MAG, this was based off selecting papers with the highest citation counts.
- In OpenAlex, as requested, this has been changed to an initial basic heuristic that favours the availability of RoR ids.
- Specifically, it looks at all listed RoR id count associated with each paper.  The paper with the highest RoR id count is the preferred paper to the multiple doi problem.  Tie breaks are resolved with arbitrary selection.

Testing:
- A new synthetic dataset creation function is introduced for OpenAlex data.
- Add tests for openalex table creation in observatory(_intermediate) datasets and for final output checking.
- Changed the doi sub field checking to allow for 2 dp precision.  This is in response to a rare incidence where a 'percentage' field failed because the values differed by 0.01.  This is likely an indication of subtle rounding differences between the SQL scripts and the Python testing environment.

cameronneylon:
I don't think I have anything substantive to add. All the changes make sense and I can see the logic of the changes. I will defer to others for the technical perspective tho!

jdddog (approved):
It looks good to me Tuan.

PR #88, opened 2022-02-22
OpenAlex fixes

aroelo:
- Add aws credentials to get transfer manifest file.
- Fix gcloud error by adding cloudsdk_python environment variable.


aroelo:
@jdddog, I hope it's alright I'll just merge this without your review, I've ran the workflow with these fixes last week. 
Otherwise OpenAlex will fail on the remote worker in the weekend.

PR #89, opened 2022-02-22
Save JSON index for the filtering API and COKI Open Access Dataset release

jdddog:
* Save a JSON index designed for the Cloudflare Worker filtering API. The file includes all countries and institutions and only saves the fields that are necessary to display in the tables and for filtering. The same code has also been used to save country.json and institution.json, which are currently loaded by the client and used to populate the tables.
* Updates for releasing the open source version of the dataset:
  *  Save the open access version of the dataset, a zip file with country.jsonl, institution.jsonl and README.
  * Changed min_year and max_year to start_year and end_year.
  * Changed timeseries to years.
  * Incremented dataset version number so that these changes won't break the web application.
* Removed code that isn't currently used, e.g. Subject, Collaborator.
* Removed parquet saving as this is no longer necessary.

### DIFF
--- a/academic_observatory_workflows/workflows/oa_web_workflow.py
+++ b/academic_observatory_workflows/workflows/oa_web_workflow.py
@@ -305,6 +305,13 @@ def save_json(path: str, data: Union[Dict, List]):
         json.dump(data, f, separators=(",", ":"))
 
 
+def val_empty(val):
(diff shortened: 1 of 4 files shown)


aroelo (dismissed):
Looks good @jdddog :-)

PR #90, opened 2022-03-01
Fix error getting wiki description oa_web_workflow

aroelo:
I have updated the way that the page titles returned by the Wikipedia API are returned match the page titles that are originally retrieved from the wikipedia_url.

This should fix the KeyError that occurred in the JIRA ticket described here https://curtinic.atlassian.net/browse/INF-353


jdddog (approved):
Thanks for the fix Aniek

PR #91, opened 2022-03-07
Refactor to use new country and region table

jdddog:
Refactor to use new country and region table


docs/tables/country.md
```diff
@@ -0,0 +1,12 @@
+# Country
+The country table is made by COKI, it contains country data, including each country's: ISO 3166 name, Wikipedia name,
+ISO 3166 two letter alpha2 code, ISO 3166 three letter alpha3 code, region name, subregion name, Wikidata QID and
+Wikipedia URL.
```
aroelo, line 4: Would it be possible to add the code or more details on where we get these values from?
E.g. if this table accidentally gets deleted, how can we recreate it?

academic_observatory_workflows/database/sql/create_ror.sql.jinja2
```diff
@@ -15,7 +15,7 @@
 # Author: Richard Hosking, James Diprose #}
 
 SELECT 
-  rors.*,
-  STRUCT(regions) as iso3166
+  rors.* EXCEPT (country),
+  country
 FROM `{{ project_id }}.{{ ror.dataset_id }}.ror{{ ror.release_date.strftime('%Y%m%d') }}` as rors
-LEFT JOIN `{{ project_id }}.{{ iso.dataset_id }}.iso3166_countries_and_regions` as regions ON (rors.country.country_code = regions.alpha2)
+LEFT JOIN `{{ project_id }}.{{ settings.dataset_id }}.country` as country ON (rors.country.country_code = country.alpha2)
```
aroelo, line 21: Apologies if you have tested this, but will it be able to still find the `rors.country.country_code` field? Because above you select `rors.* EXCEPT (country)`

aroelo (commented):
Hi @jdddog, it looks good, just two small questions.

academic_observatory_workflows/database/sql/create_ror.sql.jinja2
jdddog: Yeah, it just means that the `rors.country` field is replaced entirely by the `country` row that it is joined with. It happens after the join.

docs/tables/country.md
jdddog: It is just a CSV file that I manually created. 

I've added a CSV and JSON Lines version in the [academic_observatory_workflows/fixtures/doi](https://github.com/The-Academic-Observatory/academic-observatory-workflows/tree/feature/countries_and_regions/academic_observatory_workflows/fixtures/doi) folder and it is also loaded in BigQuery.

I've also specified a bit more information about where the region and subregion names come from.

aroelo: Nice, thanks!

aroelo (approved):

PR #92, opened 2022-03-08
Fix transform for cases where the transformed field does not exist.

aroelo:
Fixes `KeyError: 'abstract_inverted_index'`


jdddog (approved):

PR #94, opened 2022-03-20
Update file regex

jdddog:
ROR changed the compressed file name from YYYY-MM-DD-ror-data.json to v1.0-YYYY-MM-DD-ror-data.json, so file regex didn't find the correct file.


aroelo (approved):
Thanks for the fix @jdddog

PR #100, opened 2022-04-05
Fix error for Orcid telescope

aroelo:
Use gcloud fix for gsutil command as well. gsutil ran without issues until last week, using the same fix as for the gcloud command solves the problem.


jdddog (approved):

PR #103, opened 2022-04-11
Fix gsutil bug for OpenAlex

aroelo:


jdddog (dismissed):
Thanks Aniek, looks good to me.

PR #104, opened 2022-04-12
Feature/oa web acronyms

jdddog:
Export ROR acronyms with search index and remove deprecated code.


aroelo (approved):
Thanks @jdddog, will be great to have the acronyms in :-)

PR #106, opened 2022-04-12
Fix wiki descriptions

aroelo:
Fix issues that popped up with institution descriptions after setting the inclusion threshold to 100 instead of 1000.
- Fix bug when removing text between brackets if description starts with a bracket
- Fix bug where redirects from the wikipedia api point to each other


jdddog (approved):
Thanks Aniek, looks good to me.

PR #107, opened 2022-04-19
Remove largest remainder rounding

jdddog:
Move finding the largest remainder for groups of values that add up to 100% to the web app client.


aroelo (approved):
Thanks @jdddog, looks good!

PR #109, opened 2022-04-20
Update create_bigquery_table_from_query call

aroelo:
Remove 'output_cluster' argument, because signature of function changed in https://github.com/The-Academic-Observatory/observatory-platform/pull/551


jdddog (approved):
Thanks Aniek, looks good.

PR #110, opened 2022-04-21
Publish new version of dataset to Zenodo automatically

jdddog:


academic_observatory_workflows/workflows/oa_web_workflow.py
```diff
@@ -1237,6 +1353,8 @@ def __init__(
         ror_dataset_id: str = "ror",
         settings_dataset_id: str = "settings",
         version: str = "v3",
+        conceptrecid: int = 6399462,
+        zenodo_host: str = "https://zenodo.org",
```
aroelo, line 1357: Could you add descriptions for these params to the docstring just below? It is a little bit unclear what the 'conceptrecid' is

aroelo (changes requested):
Hi @jdddog, thanks it looks good.
I have one small comment on the added attributes to the OaWebWorkflow class.
Also, because the Zenodo class is mocked during the unit test for the complete workflow, it is not covered by any tests. It might be useful to add separate unit tests for this class.

academic_observatory_workflows/workflows/oa_web_workflow.py
jdddog: Thanks Aniek, yeah sure.

aroelo (approved):
Thanks @jdddog!

PR #111, opened 2022-04-22
Rename 'metrics_green_only' to 'metrics_percent_green_only'

aroelo:
Rename 'metrics_green_only' to 'metrics_percent_green_only' for data export tables, this matches with the other fields and the corresponding field in the elasticsearch mapping.

This should fix the elastic import error, because there is no 'metrics_green_only' field in the elasticsearch mappings, it had to guess the value type and this caused an error.


jdddog (approved):

PR #112, opened 2022-04-27
Fix/zenodo date version

jdddog:
Update the Zenodo published_date and version tag when making a new draft of the dataset.


aroelo (approved):
Thanks @jdddog !

PR #115, opened 2022-05-16
Unpaywall: use merged daily feeds to update

tuanchien:
Use merge_update_files to fix doi duplication issue


jdddog (approved):
Thanks Tuan, that looks good.

PR #116, opened 2022-05-25
Update OpenAlex

aroelo:
Fix to update the records correctly:
* Create a dataset release for each OpenAlex entity.
* Set the end date of each dataset release to the last update date of that entity. This update date is retrieved from the OpenAlex manifest records.
* Use the dataset release end date when deciding which files need to be downloaded from the OpenAlex s3 bucket. 

Use BigQuery snapshots, for each entity:
* Append new records into an ingestion time partitioned table, this is the 'main' table.
* If there are multiple records with the same id, then delete the older ones, this is determined by the partition date of the record.
* Create a BQ snapshot of the 'main' table each run after the table has been updated, each snapshot is stored as a sharded table.

### DIFF
--- a/academic_observatory_workflows/workflows/openalex_telescope.py
+++ b/academic_observatory_workflows/workflows/openalex_telescope.py
@@ -37,7 +37,11 @@ from google.cloud import bigquery, storage
 
 from academic_observatory_workflows.config import schema_folder as default_schema_folder
 from observatory.platform.utils.airflow_utils import AirflowVars
-from observatory.platform.utils.gc_utils import aws_to_google_cloud_storage_transfer, upload_file_to_cloud_storage
+from observatory.platform.utils.gc_utils import (
+    aws_to_google_cloud_storage_transfer,
+    upload_file_to_cloud_storage,
+    bq_delete_old_rows,
+)
 from observatory.platform.utils.proc_utils import wait_for_process
 from observatory.platform.utils.release_utils import (
     add_dataset_release,
@@ -46,7 +50,7 @@ from observatory.platform.utils.release_utils import (
     get_latest_dataset_release,
     is_first_release,
 )
-from observatory.platform.utils.workflow_utils import blob_name, bq_append_from_file, bq_delete_old_rows
+from observatory.platform.utils.workflow_utils import blob_name, bq_append_from_file
 from observatory.platform.workflows.stream_telescope import (
     StreamRelease,
     StreamTelescope,
--- a/academic_observatory_workflows/workflows/openalex_telescope.py
+++ b/academic_observatory_workflows/workflows/openalex_telescope.py
@@ -36,6 +36,7 @@ from dateutil import tz
 from google.cloud import bigquery, storage
 
 from academic_observatory_workflows.config import schema_folder as default_schema_folder
+from academic_observatory_workflows.api_type_ids import DatasetTypeId
 from observatory.platform.utils.airflow_utils import AirflowVars
 from observatory.platform.utils.gc_utils import (
     aws_to_google_cloud_storage_transfer,
@@ -329,7 +330,7 @@ class OpenAlexTelescope(StreamTelescope):
         airflow_conns: List = None,
         max_processes: int = os.cpu_count(),
         workflow_id: int = None,
-        dataset_type_id: str = "openalex",
+        dataset_type_id: str = DatasetTypeId.openalex,
     ):
         """Construct an OpenAlexTelescope instance.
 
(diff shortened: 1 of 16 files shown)


academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -300,13 +371,66 @@ def __init__(
...
         self.add_task(self.transform)
         self.add_task(self.upload_transformed)
-        self.add_task(self.bq_load_partition)
-        self.add_task_chain([self.bq_delete_old, self.bq_append_new, self.cleanup, self.add_new_dataset_releases])
+        self.add_task(self.bq_append_new)
+        self.add_task(self.bq_delete_old)
+        self.add_task(self.bq_create_snapshot)
+        self.add_task(self.add_new_dataset_releases)
+
+    def get_release_info(self, **kwargs) -> Tuple[pendulum.DateTime, pendulum.DateTime, bool]:
```
jdddog, line 384: I feel like we should be trying to use the get_release_info from the StreamTelescope rather than making a whole new method if possible. 

For OpenAlex, there is a top level DatasetType called OpenAlex which should have the correct end date set. This line `datasets = get_datasets(workflow_id=self.workflow_id)` in StreamTelescope would need to filter the datasets by a dataset_type.type_id, as there can be more than one dataset type. The type_id would need to be set in the constructor of each workflow.

Then `start_date = pendulum.instance(latest_release.created).start_of("day")` can be set back to `start_date = pendulum.instance(latest_release.end_date).start_of("day")`.

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -300,13 +371,66 @@ def __init__(
...
+            # When first release, set start date to the start of the DAG
+            # Can we get away with using data_interval_start instead?
+            start_date = pendulum.instance(kwargs["dag"].default_args["start_date"]).start_of("day")
+        else:
+            datasets = get_datasets(workflow_id=self.workflow_id)
+            releases = get_dataset_releases(dataset_id=datasets[0].id)
+            latest_release = get_latest_dataset_release(releases=releases)
+            start_date = pendulum.instance(latest_release.created).start_of("day")  # custom start date
+
+        end_date = pendulum.today(tz="UTC").start_of("day")
```
jdddog, line 406: Can the end date not be the same as get_release_info from StreamTelescope? end_date = pendulum.instance(kwargs["data_interval_end"])

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -300,13 +371,66 @@ def __init__(
...
+        necessary to create the new dataset releases.
+
+        :param release: Release passed to workflow tasks.
+        :param kwargs: kwargs passed to PythonOperator functions.
+        """
+
+        if self.workflow_id is None:
+            raise Exception("workflow_id must be set")
+
+        update_dates = release.get_update_date()
```
jdddog, line 423: You can also add a new dataset release for the top level OpenAlex dataset / dataset type

jdddog (changes requested):
Thanks Aniek, it looks good, I have some suggestions about using more of the code from the StreamTelescope and not relying on names from the API.

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -98,34 +125,48 @@ def write_transfer_manifest(self):
...
         :return: The number of updated entities.
         """
+        dt = datetime.datetime(1, 1, 1, tzinfo=tz.UTC)
+        entities = {"authors": dt, "concepts": dt, "institutions": dt, "venues": dt, "works": dt}
+
+        # Get last release date for each entity
+        if not self.first_release:
+            datasets = get_datasets(workflow_id=self.workflow_id)
+            for dataset in datasets:
+                entity = dataset.name.split()[1].lower() + "s"
```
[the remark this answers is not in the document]
aroelo: Ah okay, thanks. I did not realise that the type_id would stay the same I will change it to use this instead!

academic_observatory_workflows/workflows/openalex_telescope.py
aroelo: So you mean I would update this method for the stream telescope to always use type_id 0? (or 1, not sure whether we're starting at 0 or 1)

academic_observatory_workflows/workflows/openalex_telescope.py
aroelo: Ah yeah that's a good point, didn't think of that

academic_observatory_workflows/workflows/openalex_telescope.py
aroelo: Ah never mind, the dataset_type.type_id is actually a string I see now.

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -259,6 +260,7 @@ def __init__(
         airflow_conns: List = None,
         max_processes: int = min(32, os.cpu_count() + 4),
         workflow_id: int = None,
```
jdddog, line 262: Same question as above.

jdddog (commented):
Thanks Aniek, it looks good to me. I just have some questions about the other telescopes, see below.

jdddog (approved):
Thanks Aniek!

PR #117, opened 2022-06-01
Update fieldname "metrics_citations_per_cited_output"

aroelo:
Fixes error in elastic_import_observatory DAG:
`mapper [metrics_citations_per_cited_output] cannot be changed from type [long] to [float]'}`

Similar to a previous issue, the typing of the field 'metrics_citations_per_cited_output' was interpolated from the data instead of from the mapping file, because this field was not defined in the mapping file.

There was a 'metrics_citations_per_cited_outputs' field in ao-metrics-mappings.json.jinja2, which probably should have been 'metrics_citations_per_cited_output' instead.


jdddog (approved):
Thanks for fixing it Aniek!

PR #119, opened 2022-06-08
Fix OpenAlex prefixes limit

aroelo:
Split the prefixes that are passed to the transfer job config into chunks of 1000, this is the maximum limit for prefixes.
Ideally, the prefixes are not used at all, but instead the transfer manifest file is used. However, this is still not working, see https://issuetracker.google.com/issues/216057461.


aroelo:
Closing, because the transfer manifest file can be used instead of prefixes https://github.com/The-Academic-Observatory/academic-observatory-workflows/pull/121

PR #121, opened 2022-06-13
Update OpenAlex to use transfer manifest file

aroelo:
Update telescope to use the manifest file for the transfer job to transfer OpenAlex files from AWS to GCP.
Without updating any of the permissions that are setup/described in the docs this now seems to work, the issue here is fixed https://issuetracker.google.com/issues/216057461. 

This also means that this PR can be closed, because the prefixes are not used anymore. https://github.com/The-Academic-Observatory/academic-observatory-workflows/pull/119


jdddog (approved):

PR #123, opened 2022-06-17
Add dag tag

tuanchien:
Requires https://github.com/The-Academic-Observatory/observatory-platform/pull/571

### DIFF
--- a/academic_observatory_workflows/dag_tag.py
+++ b/academic_observatory_workflows/dag_tag.py
@@ -1,0 +1,4 @@
+class Tag:
+    """DAG tag."""
+
+    academic_observatory = "academic-observatory"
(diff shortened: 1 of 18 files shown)


jdddog (changes requested):
Thanks Tuan, these changes look good.

We should also add the `academic-observatory` TAG to: elastic_import_observatory. It will need to be a parameter passed to the workflow in the DAG file.

Could we add the `observatory-platform` tag to vm_create and vm_destroy too? That would have to be done in the main

tuanchien:
Elastic/vm tagging changes made.  Requires merging https://github.com/The-Academic-Observatory/observatory-platform/pull/572 first

jdddog (approved):
Thanks Tuan, it looks good to me.

PR #126, opened 2022-07-14
Compute histograms and change country and institution inclusion thresholds

jdddog:
This PR creates a set of histogram data for n_outputs, n_outputs_open and p_outputs_open (which are used in the range sliders on the website) and also lowers the thresholds for country and institution inclusion. The data in n_outputs_open and p_outputs_open is transformed with log base 10 to remove skew, before computing the histograms.


keegansmith21 (approved):
Looks good, thanks Jamie!

PR #127, opened 2022-07-22
OpenAlex: re-enable error checks on bq load

tuanchien:
- Re-enables error on bq load.
- Removed old table schemas.
- Renames newer table schemas to use the old date.
- Update unit tests.
- Schedule interval is left as @weekly. OpenAlex have indicated the update frequency is currently monthly.

Once merged, we need to schedule a time to notify impacted users that the old  OpenAlex data is being deleted. New tables will be cleanly re-populated from a snapshot.

The most recent OpenAlex snapshot should be 2022-07-09 which supposedly fixes the invalid jsonl problem and possibly the duplicate data issue (we need to confirm).


academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -365,7 +365,7 @@ def __init__(
             airflow_conns=airflow_conns,
             workflow_id=workflow_id,
             dataset_type_id=dataset_type_id,
-            load_bigquery_table_kwargs={"ignore_unknown_values": True},
+            load_bigquery_table_kwargs={"ignore_unknown_values": False},
```
jdddog, line 368: I would leave this as True, because it means that when there are new fields BigQuery will ignore them rather than failing and us having to update the schema before it loads. I think this workflow would need to be run from the start to update the schema and add new fields.

https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json

"(Optional) Indicates whether BigQuery should allow extra values that are not represented in the table schema. If true, the extra values are ignored. If false, records with extra columns are treated as bad records, and if there are too many bad records, an invalid error is returned in the job result. The default value is false. The `sourceFormat` property determines what BigQuery treats as an extra value: CSV: trailing columns, JSON: named values that don't match any column names."

jdddog (changes requested):
Thanks Tuan, it looks good, just a comment about the ignore_unknown_values parameter.

PR #128, opened 2022-08-03
Closes COK-322 Added All Access Type Counts to Generic and Discipline Exports

cameronneylon:
The following access types are added to the export queries for disciplines and the generic relations query and both related mappings.

The following access types are added:

* num_gold_just_doaj_outputs
* num_hybrid_outputs
* num_bronze_outputs
* num_green_only_outputs

All of these are added to the relevant mappings and mapped to longs


jdddog (approved):
Thanks Cameron, it looks good to me.

PR #130, opened 2022-08-17
INF-521: Remove DOI table copy to dashboards dataset

alexmassen-hane:
Removed the tasks that copy the DOI table to the dashboards. Will still need to manually remove the table from public-dashboards.coki_dashboards after this has been merged.


alexmassen-hane:
Forgot to add changes to the unit tests for the dag structure and remove the calls for the copy to dashboards.

jdddog (changes requested):
Thanks for the PR Alex. We just want to stop the DOI table from being copied over, not all of the tables and views. To do this it should be sufficient to remove `+ ["doi"]` from the line `table_ids = [agg.table_id for agg in DoiWorkflow.AGGREGATIONS] + ["doi"]` and same within the tests.

PR #133, opened 2022-08-23
OA web workflow other platform open updates

jdddog:
* Added Other Platform Open categories and repositories. 
  * Using new fields from aggregate tables, so I've been able to remove extra calculations that used to be done in this workflow.     
  * Updated the unit tests to take these extra fields into account.
* Refactored the workflow so that it is easier to run locally and moved most code out of the release class into the workflow as this is easier to understand.


keegansmith21 (approved):
Thanks Jamie, looks good. I especially like the simplification of the release class!

PR #134, opened 2022-08-24
OpenAlex: set start_date to end_date of recent release

tuanchien:


jdddog (approved):
Thanks Tuan, it looks good.

PR #137, opened 2022-08-30
INF-525: Removing MAG and GRID telescopes, workflows and unit tests

alexmassen-hane:
Removed MAG and GRID telescopes, dag workflows and unit tests as they are not being used anymore. 
No special functions or requirements to remove as they use standard libraries from Airflow or other shared packages from other telescopes.


jdddog (approved):
Thanks Alex, it looks good to me.

PR #140, opened 2022-09-26
Migrate aggregate/export doi metrics to openalex

tuanchien:


jdddog (approved):
Thanks Tuan, the changes look good to me.

PR #141, opened 2022-09-30
INF-557: Removed authors from the list of aggregations

alexmassen-hane:
To save space on Elastisearch, it is required that the authors aggregation table be removed from the list of aggregations being created by the DOI Workflow.

Created a seperate function in the DOI Workflow to remove one or a number of aggregations from a list of aggregation objects. The original list of aggregations are copied and the function returns a list of aggregations with the desired aggregations are removed. Aggregations to remove are matched on the "table_id".

The export for Elastic now depends on a list of aggregations with the "author" aggregation removed. No other parts of the workflow is affected. 

Unit test for the DOI Workflow have been modified for the removal of the "export_author" task.

### DIFF
--- a/academic_observatory_workflows/dags/doi_workflow.py
+++ b/academic_observatory_workflows/dags/doi_workflow.py
@@ -33,7 +33,6 @@ Each release is saved to the following BigQuery tables:
 
 Every week the following tables are overwritten for visualisation in the Data Studio dashboards:
     <project_id>.coki_dashboards.country
-    <project_id>.coki_dashboards.doi
     <project_id>.coki_dashboards.funder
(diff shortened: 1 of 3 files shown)


jdddog (commented):
Hey Alex, it looks good. I would just change the list of items in remove_aggregations to a set rather than a list. It is not going to make a difference at all in reality because the list of items is so small, however, it is good to be aware of as it is much more efficient to check if an item exists in a set than a list.

alexmassen-hane:
I have made the change from a List to a Set. Have also fixed the unit test with Jamie's assistance.

jdddog (approved):
Thanks Alex, it looks good to me.

PR #142, opened 2022-10-14
OA Web Open Alex

jdddog:
Updated the OA Web Workflow to better support the volume of institutional data (87K institutions) present in the Open Alex data, including:
* The queries are split into two parts: the index queries fetch the unique countries and institutions and metadata, whilst the data query fetches the timeseries data. This makes the Pandas processing use less memory because the timeseries data can be processed without each entities metadata repeated for each timepoint in the dataframe.
* The `download_twitter_cards` task renamed to `download_cached_assets`, which now also downloads a zip file of logos so that we don't need to fetch all of the logos from clearbit every time, which takes a long time now that we have 87K institutions.
* Split of the previous `transform` task into multiple tasks so that if a task fails the same tasks don't need to be repeated.
  * preprocess_data: pre-process data fetched by the data query.
  * build_indexes: makes indexes from data fetched by the index queries.
  * download_logos: downloads logos. Now it logs its progress.
  * download_wiki_descriptions: downloads wikipedia descriptions. Now it logs its progress.
  * build_datasets: builds the final datasets required for the website and Zenodo.

Other changes:
* Simplified code where possible.
* Moved supporting libraries into their own modules so that there is less code on oa_web_workflow.py. These new modules are in clearbit.py, github.py, wikipedia.py and zenodo.py.
* Renamed Bonaire flag from BQ-BO.svg to BQ.svg so it is picked up by the workflow.
* Added Kosovo to country table.


alexmassen-hane (approved):
Looks good. I like how tasks have been separated out and will be easier to track down issues in the future. Keen to see how the data looks after the workflow is run.

jdddog:
> Looks good. I like how tasks have been separated out and will be easier to track down issues in the future. Keen to see how the data looks after the workflow is run.

Thanks Alex

PR #143, opened 2022-10-20
Check for invalid lat lng values

jdddog:
The ROR record https://ror.org/05g8baj95 has an invalid longitude of -7402092.0, which breaks importing to Elasticsearch. I've logged a report at ROR, however, in case we get invalid coordinates in the future, I've added a check to make sure that the address coordinates are valid and if they aren't print a warning and set them to None.

Move some of the code out of the release to make reading more straightforward as we are doing in the newer workflows.


keegansmith21 (commented):
Looks good, thanks Jamie!

keegansmith21 (approved):

PR #144, opened 2022-10-25
DOI Workflow: add genealogical institutions

jdddog:
This PR adjusts the DOI workflow to produce aggregate statistics based on direct institutions (author institutions) and ancestral institutions (the ancestral institutions of the direct author institutions). This should produce institutional statistics more in line with MAG. 
* Adds a task to the DOI workflow which creates an index table that maps ROR ids to all ancestor ROR ids using the ROR parent child relationships.
* Modify DOI table coki struct to list author institution ROR ids and genealogical institutions ROR ids (author institutions + their ancestors).
* Modify create_doi.sql.jinja2 to produce aggregations based on genealogical institutions rather than author institutions.


alexmassen-hane (approved):
Looks good Jamie!

PR #150, opened 2022-11-30
INF-572 : query formatting / linting and minor refactoring

JulianTonti:
* export_unique_list - formatting only
* export_relations - formatting only
* export_output_types - formatting only
* export_metrics - formatting only
* export_events - formatting only
* export_dois - formatting and one suggestion
* export_disciplines - formatting only
* export_access_types - formatting only
* create_wos - formatting only
* create_unpaywall -
  * refactored repositories sub-query
  * replaced binary CASEs with IFs
  * replaced other_platform_categories CASE with COUNTIF
* create_unpaywall_repo_names - switched to IFNULL
* create_scopus - formatting only
* create_ror - formatting only
* create_orcid - formatting only
* create_openalex - formatting only
* create_open_citations - fully refactored query
* create_mag - formatting, switched one IF for IFNULL
* create_doi -
  * switched IFs for IFNULLs
  * one query on count field
  * added consistent use of CASTs
  * formatting
* create_crossref_fundref - formatting only
* create_crossref_events - formatting only
* create_book - refactored JavaScript UDF and added tests
* create_aggregate - formatting only
* comparison_view - formatting only

### DIFF
--- a/academic_observatory_workflows/database/sql/comparison_view.sql.jinja2
+++ b/academic_observatory_workflows/database/sql/comparison_view.sql.jinja2
@@ -1,26 +1,27 @@
-{# Copyright 2020 Curtin University
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#   http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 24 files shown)


jdddog (changes requested):
Thanks for tidying up the SQL Julian, it is much clearer.

What tool are you using for formatting the SQL? It would be good to standardise our approach so that when others are working on the same files we keep the same formatting.

The unit tests for the DOI Workflow are failing which need to be fixed before we merge it.

PR #152, opened 2022-12-16
Feature/oa web d1 database

jdddog:
This PR just makes a few more small changes to the output data structures in preparation of using Cloudflare D1 and showing OpenAlex data on the website. It also changes what institutions are included in the final dataset, including all institutions in the last release of the dashboard + any institution from the OpenAlex dataset with 700 publications or more. We will include them all after using D1 for the backend.

It looks like a lot of changes, however, that is because I've removed the country flags, these are instead kept on a Google Cloud Storage bucket, along with other files that can be used to override logos that are automatically downloaded, or cached.


keegansmith21 (approved):
Thanks Jamie, this looks good! :+1:

PR #153, opened 2023-01-17
Updated regex pattern match

keegansmith21:
This PR alters the ROR extract_files regex pattern match. 
The reason being, a recent update (by ROR) changed the version number to 1.17.1. The old regex did not pick this up because it is not able to match on version numbers outside of the format [major].[minor]. The updated regex looks for any number of sub-versions. (e.g. 1.17.1.2.3.4.5)


alexmassen-hane (approved):
Thanks Keegan. Looks good.

PR #158, opened 2023-02-27
Feature/workflow config file

jdddog:


academic_observatory_workflows/database/sql/create_aggregate.sql.jinja2
```diff
@@ -801,7 +801,7 @@ WITH tmp_disciplines AS
             FROM UNNEST(openalex.concepts) as fields where fields.level = 0)) )) as level0
       ) as disciplines
   FROM
-    `{{ project_id }}.{{ dataset_id }}.doi{{ release_date.strftime('%Y%m%d') }}` as dois,
+    `{{ project_id }}.{{ dataset_id }}.doi{{ snapshot_date.strftime('%Y%m%d') }}` as dois,
```
keegansmith21, line 804: For cases like this in the Jinja template, I wonder if it's better to just supply the fully qualified table ID? It makes the template cleaner and leaves any and all logic to python. What do you think?

keegansmith21: Update: I see some other templates have been implemented this way ^ Is there a case for not doing the same for this one and its siblings?

academic_observatory_workflows/workflows/crossref_events_telescope.py
```diff
@@ -12,383 +12,859 @@
...
-    return event_type, date
+    events = []
+    period = pendulum.period(start_date, end_date)
+    for day in period.range("days"):
+        if day != end_date:
+            events.append(DayRequest(day, mailto))
+    return events
+
 
+def download_events(request: EventRequest, download_folder: str, n_rows: int):
```
keegansmith21, line 737: Codecov says that the first two 'if' statements aren't covered in tests. It might be worth making sure they're included as the logic flow here is pretty important.

academic_observatory_workflows/workflows/crossref_events_telescope.py
```diff
@@ -12,383 +12,859 @@
...
+        self.add_task(self.bq_load_delete_table)
+        self.add_task(self.bq_delete_records)
+
+        # Add release info to API and cleanup
+        self.add_task(self.add_new_dataset_releases)
+        self.add_task(self.cleanup)
+
+        # The last task that the next DAG run's ExternalTaskSensor waits for.
+        self.add_operator(
+            DummyOperator(
```
keegansmith21, line 278: Why use a dummy here over the cleanup task?

academic_observatory_workflows/workflows/crossref_events_telescope.py
```diff
@@ -12,383 +12,859 @@
...
+        query_params["until-updated-date"] = end_date.format(DATE_FORMAT)
+
+    if mailto is not None:
+        query_params["mailto"] = mailto
+    if rows is not None:
+        query_params["rows"] = rows
+    if cursor is not None:
+        query_params["cursor"] = cursor
+
+    req = requests.models.PreparedRequest()
```
keegansmith21, line 676: This looks great, I didn't know the requests library had something like this. Will a request made in this manner always work (ie. correctly format the url) for a REST API?

academic_observatory_workflows/workflows/crossref_events_telescope.py
```diff
@@ -12,383 +12,859 @@
...
+
+
+def crossref_events_limiter(calls_per_second: int = 10):
+    """Function to throttle the calls to the Crossref Events API"""
+
+    identifier = "crossref_events_limiter"
+    item = RateLimitItemPerSecond(calls_per_second)  # 10 per second
+
+    while True:
+        if not moving_window.test(item, identifier):
```
keegansmith21, line 832: Should moving_window be capitalised to indicate a global variable?

academic_observatory_workflows/workflows/crossref_fundref_telescope.py
```diff
@@ -16,242 +16,131 @@
...
-        create_pool(pool_name, num_slots, description)
+        create_pool(gitlab_pool_name, gitlab_pool_slots, gitlab_pool_description)
 
         self.add_setup_task(self.check_dependencies)
-        self.add_setup_task(self.get_release_info, pool=pool_name)
-        self.add_task(self.download, pool=pool_name)
+        self.add_setup_task(self.get_release_info, pool=gitlab_pool_name)
+        self.add_task(self.download, pool=gitlab_pool_name)
         self.add_task(self.upload_downloaded)
         self.add_task(self.extract)
```
keegansmith21, line 138: Is the extract task necessary for fundref? The data is pretty small?

academic_observatory_workflows/workflows/doi_workflow.py
```diff
@@ -342,15 +341,16 @@ def fetch_ror_affiliations(repository_institution: str, num_retries: int = 3) ->
...
-    table_shard_dates = select_table_shard_dates(project_id, dataset_id, table_name, release_date)
+    logging.info(f"get_snapshot_date {project_id}.{dataset_id}.{table_id} {snapshot_date}")
+    table_id = bq_table_id(project_id, dataset_id, table_id)
+    table_shard_dates = bq_select_table_shard_dates(table_id=table_id, end_date=snapshot_date)
     if len(table_shard_dates):
         shard_date = table_shard_dates[0]
     else:
         raise AirflowException(
-            f"{project_id}.{dataset_id}.{table_name} " f"with a table shard date <= {release_date} not found"
+            f"{project_id}.{dataset_id}.{table_id} " f"with a table shard date <= {snapshot_date} not found"
```
keegansmith21, line 353: This should just be {table_id} without the project id and dataset

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -21,590 +21,923 @@
...
+        self.add_operator(
+            DummyOperator(
+                task_id=external_task_id,
+            )
+        )
 
-        :param release: Release passed to workflow tasks.
-        :param kwargs: kwargs passed to PythonOperator functions.
+    @property
+    def aws_key(self) -> Tuple[str, str]:
```
keegansmith21, line 412: I like this solution for streamlining airflow connections! Perhaps we should incorporate something similar with other telescopes in the future?

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -21,590 +21,923 @@
...
+
+        # Download concepts, institutions and works which need to be pre-processed
+        # Gsutil is used instead of the standard Google Cloud Python library, because it is faster at downloading files
+        # than the Google Cloud Python library.
+        for entity_name, transform in self.entities:
+            if transform:
+                output_folder = "{{ release.download_folder }}/data/" + entity_name + "/"
+                bucket_path = "{{ release.gcs_openalex_data_uri }}data/" + entity_name + "/*"
+                self.add_operator(
+                    WorkflowBashOperator(
```
keegansmith21, line 377: Just a cosmetic comment - this might look nicer in its own function. Up to you.

academic_observatory_workflows/workflows/unpaywall_telescope.py
```diff
@@ -12,332 +12,717 @@
...
+        table_description: str = "Unpaywall Data Feed: https://unpaywall.org/products/data-feed",
+        primary_key: str = "doi",
+        snapshot_expiry_days: int = 7,
+        http_header: str = None,
+        unpaywall_conn_id: str = "unpaywall",
+        observatory_api_conn_id: str = AirflowConns.OBSERVATORY_API,
+        start_date: pendulum.DateTime = pendulum.datetime(2021, 7, 2),
+        schedule_interval: str = "@daily",
+    ):
+        """Unpaywall Data Feed telescope."""
```
keegansmith21, line 213: Just missing the parameters for this docstring

academic_observatory_workflows/workflows/unpaywall_telescope.py
```diff
@@ -12,332 +12,717 @@
...
-from academic_observatory_workflows.dag_tag import Tag
 
+# See https://unpaywall.org/products/data-feed for details of available APIs
+SNAPSHOT_URL = "https://api.unpaywall.org/feed/snapshot"
+CHANGEFILES_URL = "https://api.unpaywall.org/feed/changefiles"
+CHANGEFILES_DOWNLOAD_URL = "https://api.unpaywall.org/daily-feed/changefile"
+
+
+class Changefile:
+    def __init__(self, filename: str, changefile_date: pendulum.DateTime, changefile_release: ChangefileRelease = None):
```
keegansmith21, line 78: A docstring here wouldn't go amiss

keegansmith21 (changes requested):
Thank Jamie, the changes all look really good. Just a few very minor comments that would be good to take a look at before merging.

academic_observatory_workflows/workflows/crossref_events_telescope.py
jdddog: Yeah that makes sense.

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -21,590 +21,923 @@
...
+    WorkflowBashOperator,
+    ChangefileRelease,
+    set_task_state,
+    cleanup,
 )
-from academic_observatory_workflows.dag_tag import Tag
 
 
-class OpenAlexRelease(StreamRelease):
+class OpenAlexEntity:
```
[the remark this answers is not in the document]
jdddog: Yeah sure

academic_observatory_workflows/database/sql/create_aggregate.sql.jinja2
jdddog: Yeah it would be better to use the approach that you suggest wherever possible.

Some of those templates only specify a project_id and dataset_id and then there are multiple table names referenced off them, so those instances were a can of worms to sort out.

We could have another crack when we refactor the DOI table?

academic_observatory_workflows/workflows/crossref_events_telescope.py
jdddog: I guess it isn't strictly necessary, but it makes it easier to test the workflows because the cleanup task can be commented out during testing.

academic_observatory_workflows/workflows/crossref_events_telescope.py
jdddog: Thanks. Yeah it should do, we could use this in other places.

academic_observatory_workflows/workflows/crossref_fundref_telescope.py
jdddog: It could be done in memory. Although extracting it on disk does mean that you can inspect the extracted files if need be.

academic_observatory_workflows/workflows/openalex_telescope.py
jdddog: Thanks. Would you suggest consolidating these functions in the future?

get_airflow_connection_login
get_airflow_connection_password
get_airflow_connection_url

academic_observatory_workflows/workflows/crossref_events_telescope.py
jdddog: Yeah sure

academic_observatory_workflows/workflows/crossref_events_telescope.py
keegansmith21: Ah I see. Fair enough

academic_observatory_workflows/workflows/openalex_telescope.py
keegansmith21: Sure, do you think it'd be better to combine them and choose the output accordingly for each use?

academic_observatory_workflows/database/sql/create_aggregate.sql.jinja2
keegansmith21: Yeah I think so. Is there a Jira ticket for the DOI refactor? If not I can make one and mention this so it's not forgotten.

academic_observatory_workflows/workflows/crossref_events_telescope.py
jdddog: I've added in another test as well to check `fetch_events` as well, with the real endpoints.

academic_observatory_workflows/database/sql/create_aggregate.sql.jinja2
jdddog: Thanks that would be great. I can't see one.

academic_observatory_workflows/workflows/openalex_telescope.py
jdddog: Probably. What those functions do is assert that the right parts of the connection were specified, so there might need to be some keyword parameters to turn the checks for specific parameters on and off.

jdddog:
> Thank Jamie, the changes all look really good. Just a few very minor comments that would be good to take a look at before merging.

Thanks for your review Keegan, I've made most of the changes apart from the DOI query / dummy task ones.

keegansmith21 (dismissed):
Thanks for looking into the suggestions. Looks really good, Jamie :+1:

PR #160, opened 2023-03-01
Pubmed Telescope

alexmassen-hane:
Workflow to ingest the Pubmed Medline citation database. 

Requires the below to be merged before tests can pass.
https://github.com/The-Academic-Observatory/observatory-platform/pull/621


academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+
+
+class PubmedEntity:
+    def __init__(
+        self,
+        name: str,
+        type: str,
+        sub_key: str,
+        set_key: str,
+        PMID_location: str,
```
keegansmith21, line 226: I'd rename this pmid_location to keep consistent snake casing

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+
+
+class PubMedRelease(ChangefileRelease):
+    def __init__(
+        self,
+        *,
+        dag_id: str,
+        run_id: str,
+        cloud_workspace: CloudWorkspace,
+        bq_dataset_id: str,
```
keegansmith21, line 253: I don't think the bq_dataset_id is used by the release anywhere. It shouldn't really be tied to the release anyhow so you can probably remove this

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+
+            # Change to the baseline directory.
+            ftp_conn.cwd(self.baseline_path)
+
+            # Grab list of all baseline files to be downloaded.
+            baseline_list_ftp = ftp_conn.nlst()
+
+            # Grab metadata and path of the file.
+            for file in baseline_list_ftp:
+                if file.split(".")[-1] == "gz" in file:  # Find all the xml.gz files available from the server.
```
keegansmith21, line 481: slightly neater to do a file.endswith(".xml.gz")

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+        self.add_task(self.cleanup)
+
+        # The last task that the next DAG run's ExternalTaskSensor waits for.
+        self.add_operator(
+            DummyOperator(
+                task_id=external_task_id,
+            )
+        )
+
+    def list_changefiles_for_release(self, **kwargs):
```
keegansmith21, line 454: just needs a -> bool

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+
+                    # If md5 does not match, raise an Airflow exception.
+                    if md5hash_from_download in md5_from_pubmed_ftp:
+                        download_success = True
+                    else:
+                        logging.info(
+                            f"MD5 hash does not match the given MD5 checksum from server: {changefile.download_file_path}"
+                        )
+
+                    download_attemp_count += 1
```
keegansmith21, line 686: 'attemp' -> attempt

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+                        md5hash_from_download = hashlib.md5(data).hexdigest()
+
+                    logging.info(f"MD5 has from downloaded file: {md5hash_from_download}")
+
+                    # Need to have a download catch for the hash file as well, otherwise it can break the download loop.
+                    try:
+                        # Download corresponding md5 hash.
+                        with open(f"{download_location}.md5", "wb") as f:
+                            ftp_conn.retrbinary(f"RETR {changefile.path_on_ftp}.md5", f.write)
+                    except:
```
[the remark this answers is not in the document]
keegansmith21: https://docs.python.org/2/library/ftplib.html#ftplib.error_reply

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+
+        # Grab list of files to upload.
+        changefiles_to_upload = [changefile.download_file_path for changefile in release.changefile_list]
+
+        success = gcs_upload_files(
+            bucket_name=self.cloud_workspace.download_bucket,
+            file_paths=changefiles_to_upload,
+        )
+
+        if not success:
```
keegansmith21, line 720: Jamie made a function during the observatory platform's recent refactor that's built to streamline success checks. See here for an example: 
https://github.com/The-Academic-Observatory/oaebu-workflows/blob/30e3ed251e7d348d29ad1f774ade6f08bde0fc96/oaebu_workflows/workflows/jstor_telescope.py#L293

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+        """
+
+        ti: TaskInstance = kwargs["ti"]
+        merged_updatefiles = ti.xcom_pull(key="merged_updatefiles")
+
+        # If its the first run of the workflow, only upload the baseline table to the main table.
+        if release.is_first_run:
+            # Create the dataset if it doesn't exist already.
+            try:
+                bq_create_dataset(
```
keegansmith21, line 921: You don't need the try/except, bq_create_dataset will catch the error raised for an existing dataset

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+    # Loop through the Pubmed record metadata, pulling out additions or deletions.
+    for entity in entity_list:
+        logging.info(
+            f"Extracting {entity.set_key if entity.type == 'deletions' else entity.sub_key} records from file - {changefile.filename}"
+        )
+
+        try:
+            try:
+                data_part = [retrieve for retrieve in data_dict[entity.set_key][entity.sub_key]]
+            except:
```
keegansmith21, line 1153: What would we be excepting here? KeyError?

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1345 @@
...
+            return [self._transform_obj_data(elem) for elem in obj]
+        else:
+            return obj
+
+    def encode(self, obj):
+        transformed_obj = self._transform_obj_data(obj)
+        return super(PubMedCustomEncoder, self).encode(transformed_obj)
+
+
+def bq_delete_table(
```
keegansmith21, line 1293: Oh this is great, I was wondering why we didn't have a bq_delete table!

keegansmith21 (changes requested):
Thanks Alex, this looks pretty good after a first pass! Just a few comments - mostly minor, but would be good to mitigate the blanket try/excepts.

academic_observatory_workflows/database/schema/pubmed/article_additions_2023-01-01.json
```diff
@@ -0,0 +1,1504 @@
+[
```
jdddog, line 1: I would just call this pubmed.json and the table pubmed, since it isn't loading sharded data and we will have to re-run the workflow if we want to change the schema.

academic_observatory_workflows/database/schema/pubmed/article_deletions_2023-01-01.json
```diff
@@ -0,0 +1,14 @@
+[
```
jdddog, line 1: Could just call it deletes.json.

jdddog (changes requested):
Hey Alex

Nice work getting the PubMed telescope to this stage, the downloading, transforming and unit tests look good. My main suggestion is that it would be good to make the workflow more consistent with the other workflows that load data incrementally, so that it will be easier to understand and maintain the workflow in the future. Some more detailed comments are below.

* On the first run, the workflow downloads the latest snapshot and all of the changefiles it needs to get up to date. Don't reply on catchup or the Airflow start_date.
* PubMed recommends that when they release a new baseline each year, the new snapshot should be downloaded and reloaded. So we will need to re-run the workflow once a year. 
  *  How will we know if the baseline and updates have moved over to a new year? We could add a check into `list_changefiles_for_release` that makes sure that the baseline and changefiles match the data interval and throw an error if they don't match so that we know that we need to start the workflow again.
  * It would be good to not rely on the Airflow start date, because otherwise we will have to be careful to change it each time that we re-run the workflow.
* Split code out into tasks, rather than having many if statements and loops within tasks, which makes it harder to follow the code but also when tasks fail they have to repeat more steps than necessary on the retries. This could involve:
  * Taking the snapshot of the main table in its own task.
  * Move loading the snapshot into its own tasks, similar to the other workflows. In the future we can make it more obvious that this task is only executed on the first run, by making it skip on subsequent runs.
  * Separating out upserting and deleting into their own sets of tasks, similar to the other workflows.
  * For the tasks after transform, not using the PubmedEntity / entity_list to loop over update and delete tasks. Using the pattern from the other workflows will mean that this code gets split out into different tasks naturally. The OpenAlex workflow has a list of entities, although this is slightly different, but will still also need to be made into different tasks sometime soon for the same reasons.
* Move loading the snapshot into its own tasks, similar to the other workflows.
  * After transforming the snapshot files, upload to Google Cloud Storage, then load into BigQuery.
  * Don't need to run merge_update_files for the baseline.
  * There is code in merge_update_files that is used for the baseline, however, to simplify things we could remove the code in merge_update_files that bundles the files together and keeps them under a certain size, either outputting non-compressed files or compressed files if the parts are always going to be smallish.
* Transforming and merging update files
  * merge_update_files can then just be dedicated to the changefiles.
  * The transformation step would need to output two sets of files, one for upserts and one for deletes, which bq_upsert_records and bq_delete_records can use.
  * See the exploration of the PubMed data below, as this has implications for what the merging needs to do when merging the updates and deletes across all changefiles.
 
* Use `bq_upsert_records` and `bq_delete_records`.
  * These need to be modified to accept multiple keys when matching, so that the PMID and Version can both be passed.
  *  You would load the upsert and delete tables in separate tasks. There is no need to delete the upsert and delete tables before hand, you can just overwrite them.

## PubMed inserts, updates and deletes
I've explored the PubMed data to understand how the inserts, updates and deletes work.
The `<PMID Version="1">` attribute is used to support publishers that publish multiple versions of a work. When the documentation was written, only one journal was using the Version attribute, which was PLoS Currents. 

See these links for more details: 
* https://www.nlm.nih.gov/bsd/licensee/elements_descriptions.html#pmid
* https://www.nlm.nih.gov/bsd/licensee/elements_descriptions.html#versionid

### Baseline
Are there multiple Versions of a single PMID in the baseline? 
* Yes. The table below shows a count of how many PMIDs exist with different numbers of versions.
```
version_count	pmid_count
1	        34954194
2	        2417
3	        431
4	        61
5	        17
6	        4
8	        2
10	        1
```

So you're right that we do need to take the version into account.
## Updates
What are the delete citations deleting?
* Past records from previous days: yes. There are 10,892 deletes of past records.
* Records that are to be upserted that day: No. There are 0 deletes of records from the same day.
* Records that don't exist: yes. There are 2,293 deletes of records that don't exist in the most up to date version of the dataset.
  *  What versions are deleted that don't exist? Version 1: 2201 records, Version 2: 92 records.

Are records ever inserted or updated multiple times in a day?
* No, there is only ever once instance of a PMID + Version in a single sequence file.

Is the same record ever deleted multiple times in a day?
* No, there are 0 records deleted multiple times on the same day.

Is the same record ever deleted on more then one day?
* Yes, there are 1203 records deleted multiple times, but on different days.

I think this means (please correct me if not), when merging changefiles:
* Need to keep the latest update of a particular PMID + Version across all changefiles.
* Don't have to worry about multiple upserts from the same changefile or deletes of those upserts from the same changefile.
* Need to remove deletes that are upserted over in subsequent changefiles and take duplicate deletes of the same PMID + Version across all changefiles into account.

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+        merged_upsert_files = ti.xcom_pull(key="merged_upsert_files")
+
+        success = gcs_upload_files(
+            bucket_name=self.cloud_workspace.transform_bucket,
+            file_paths=merged_upsert_files,
+        )
+
+        set_task_state(success, kwargs["ti"].task_id, release=release)
+
+    def bq_load_upsert_table(self, release: PubMedRelease, **kwargs):
```
jdddog, line 1001: It would be good to make the upsert, delete and snapshot names consistent with other workflows:
* pubmed_backup -> pubmed_snapshotYYYYMMDD.
* delete -> pubmed_delete (has table name in front and not sharded). The table name is in front as a namespace for the delete and upsert tables when there are multiple tables. You can make the tables expire, that is a good addition.
* upsert -> pubmed_upsert (has table name in front and not sharded).

You might also to make the schema names more consistent with other workflows or this naming, e.g. deletes.json delete.json and maybe upserts.json as pubmed.json.

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+            ),
+            "delete": PubmedEntity(
+                type="delete",
+                sub_key="PMID",
+                set_key="DeleteCitation",
+                pmid_location=None,
+                table_name="deletes",
+                table_description="""DeleteCitation - Indicates one or more <PubmedArticle> or <PubmedBookArticle> that have been deleted. 
+                PMIDs in DeleteCitation will typically have been found to be duplicate citations, or citations to content 
+                that was determined to be out-of-scope for PubMed. It is possible that a PMID would appear in DeleteCitation
```
jdddog, line 414: It would be simpler to just set these directly as variables, then it is easier to refer to them throughout the workflow:

```python
        self.baseline = PubmedEntity(
            type="baseline",
            # sub_key="PubmedArticle",
            # set_key="PubmedArticleSet",
            # pmid_location="MedlineCitation",
            table_name="upserts",
            table_description="""Pubmed's PubmedArticle table - Includes all the metadata associated with a journal article citation, both the metadata to describe the published article, i.e. <MedlineCitation>, and additional metadata often pertaining to the publication's history or processing at NLM, i.e. <PubMedData>.""",
        )
        self.upsert = PubmedEntity(
            type="upsert",
            # sub_key="PubmedArticle",
            # set_key="PubmedArticleSet",
            # pmid_location="MedlineCitation",
            table_name="upserts",
            table_description="""PubmedArticle - Includes all the metadata associated with a journal article citation, both the metadata to describe the published article, i.e. <MedlineCitation>, and additional metadata often pertaining to the publication's history or processing at NLM, i.e. <PubMedData>.""",
        )
        self.delete = PubmedEntity(
            type="delete",
            # sub_key="PMID",
            # set_key="DeleteCitation",
            # pmid_location=None,
            table_name="deletes",
            table_description="""DeleteCitation - Indicates one or more <PubmedArticle> or <PubmedBookArticle> that have been deleted. PMIDs in DeleteCitation will typically have been found to be duplicate citations, or citations to content that was determined to be out-of-scope for PubMed. It is possible that a PMID would appear in DeleteCitation without having been distributed in a previous file. This would happen if the creation and deletion of the record take place on the same day.""",
        )
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+
+        # Process files in batches so that ProcessPoolExecutor doesn't deplete the system of memory
+        for i, chunk in enumerate(get_chunks(input_list=datafiles_to_transform, chunk_size=self.max_processes)):
+            with ProcessPoolExecutor(max_workers=self.max_processes) as executor:
+                logging.info(f"In chunk {i} and processing files: {chunk}")
+
+                futures = [
+                    executor.submit(transform_pubmed_xml_file_to_jsonl, datafile, [self.entity_list["baseline"]])
+                    for datafile in chunk
+                ]
```
jdddog, line 774: I think it would be clearer to explicitly pass in what files to transform (see suggestions at the end on the transforming code):

```python
futures = []
for datafile in chunk:
    input_path = datafile.download_file_path
    upsert_path = datafile.transform_file_path(entity_type=self.baseline.type)
    futures.append(executor.submit(transform_pubmed, input_path, upsert_path))
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+                logging.info(f"In chunk {i} and processing files: {chunk}")
+
+                futures = [
+                    executor.submit(
+                        transform_pubmed_xml_file_to_jsonl,
+                        datafile,
+                        [self.entity_list["upsert"], self.entity_list["delete"]],
+                    )
+                    for datafile in chunk
+                ]
```
jdddog, line 906: I think this would make it clearer what files are being input and created:
```python
futures = []
for datafile in chunk:
    input_path = datafile.download_file_path
    upsert_path = datafile.transform_file_path(entity_type=self.upsert.type)
    delete_path = datafile.transform_file_path(entity_type=self.delete.type)
    futures.append(executor.submit(transform_pubmed, input_path, upsert_path, delete_path))
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+        """
+
+        upsert_files_to_merge = [datafile for datafile in release.datafile_list if not datafile.baseline]
+
+        upsert_files = {}
+        for datafile in upsert_files_to_merge:
+            transform_file = datafile.transform_file_path(entity_type=self.entity_list["upsert"].type)
+
+            with open(transform_file, "r") as f_in:
+                incoming_data = [json.loads(line) for line in f_in]
```
jdddog, line 930: There are pre-made functions `from observatory.platform.files import load_jsonl, save_jsonl_gz, save_jsonl` that can be used to load and save a lot of the files.

```suggestion
            transform_file = datafile.transform_file_path(entity_type=self.entity_list["upsert"].type)
            incoming_data = load_jsonl(transform_file)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+            # Add newest updatefile to list of updatefiles and it's records.
+            upsert_files[transform_file] = records
+
+        merged_upsert_files = []
+        first_file_index = upsert_files_to_merge[0].file_index
+        last_file_index = upsert_files_to_merge[-1].file_index
+        for file_count, (transform_file, records_to_write) in enumerate(upsert_files.items()):
+            logging.info(f"Reading in file to pull out records - {transform_file}")
+            with open(transform_file) as f_in:
+                incoming_data = [json.loads(line) for line in f_in]
```
jdddog, line 964:
```suggestion
            incoming_data = load_jsonl(transform_file)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+            merged_output_file = datafile.merged_transform_file_path(
+                self.entity_list["upsert"].type, first_file_index, last_file_index, part_num=file_count + 1
+            )
+
+            logging.info(f"Writing {len(records_to_write)} records to file - {merged_output_file}")
+
+            # Write to compressed *.jsonl.gz as this will be uploaded to GCS and imported into BQ.
+            with gzip.open(merged_output_file, "w") as f_out:
+                for record_key in records_to_write:
+                    f_out.write(str.encode(json.dumps(records[record_key], cls=PubMedCustomEncoder) + "\n"))
```
jdddog, line 980: You could make a function `save_pubmed_jsonl` and use this in multiple places:

```suggestion
            save_pubmed_jsonl(merged_output_file, [records[record_key] for record_key in records_to_write])
```

For instance:
```python
def save_pubmed_jsonl(file_path: str, data: List[dict]):
    if file_path.endswith(".gz"):
        with gzip.open(file_path, "w") as f_out:
            for line in data:
                f_out.write(str.encode(json.dumps(line, cls=PubMedCustomEncoder) + "\n"))
    else:
        with open(file_path, "w") as f_out:
            for line in data:
                f_out.write(json.dumps(line, cls=PubMedCustomEncoder) + "\n")
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+        merged = []
+        first_file_index = delete_files_to_merge[0].file_index
+        for datafile in delete_files_to_merge:
+            transform_file = datafile.transform_file_path(entity_type=self.entity_list["delete"].type)
+
+            logging.info(f"Reading in file - {transform_file}")
+
+            # Read in delete file
+            with open(transform_file, "r") as f_in:
+                incoming_data = [json.loads(line) for line in f_in]
```
jdddog, line 1074:
```suggestion
            incoming_data = load_jsonl(transform_file)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+            merged_output_file = datafile.merged_transform_file_path(
+                self.entity_list["delete"].type, first_file_index, datafile.file_index, part_num=i + 1
+            )
+
+            logging.info(f"Writing to file - {merged_output_file}")
+
+            # Write to compressed *.jsonl.gz as this will be uploaded to GCS and imported into BQ.
+            with gzip.open(merged_output_file, "w") as f_out:
+                for line in chunk:
+                    f_out.write(str.encode(json.dumps(line, cls=PubMedCustomEncoder) + "\n"))
```
jdddog, line 1094:
```suggestion
            save_jsonl_gz(merged_output_file, chunk)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+        ), f"Unable to download {datafile.download_file_path} from PubMed's FTP server \
+                    {ftp_server_url} after {max_download_retry} tries."
+
+    # Close the FTP connection after downloading the required files.
+    ftp_conn.close()
+
+    return download_success
+
+
+def transform_pubmed_xml_file_to_jsonl(datafile: Datafile, entity_list: List[PubmedEntity]) -> Union[Datafile, bool]:
```
jdddog, line 1307: I would give it a shorter name, e.g. `transform_pubmed` and split this function out into a few different functions, e.g. as below. I think it is easier to understand what is going on just by directly setting the dictionary keys / making explicit functions to get the articles / deletes out of the data.

```python
def load_datafile(file_path: str):
    logging.info(f"Reading in file - {file_path}")

    with gzip.open(file_path, "rb") as f_in:
        # Use the BioPython package for reading in the Pubmed XML files.
        # This package also checks against its own DTD schema defined in the XML header.
        data = Entrez.read(f_in, validate=True)

    # Need pull out XML attributes from the Biopython data classes.
    data = add_attributes_to_data_from_biopython_classes(data)

    #  Remove unwanted nested list structure from the Pubmed dictionary.
    data = change_pubmed_list_structure(data)

    return data


def save_pubmed_jsonl(file_path: str, data: List[dict]):
    if file_path.endswith(".gz"):
        with gzip.open(file_path, "w") as f_out:
            for line in data:
                f_out.write(str.encode(json.dumps(line, cls=PubMedCustomEncoder) + "\n"))
    else:
        with open(file_path, "w") as f_out:
            for line in data:
                f_out.write(json.dumps(line, cls=PubMedCustomEncoder) + "\n")


def parse_articles(data: Dict):
    try:
        return data["PubmedArticle"]
    except KeyError:
        logging.info(f"No PubmedArticle records in file")
        return []


def parse_deletes(data: Dict):
    try:
         return data["DeleteCitation"]["PMID"]
    except KeyError:
        logging.info(f"No DeleteCitation.PMID records in file")
        return []


def transform_pubmed(input_path: str, upsert_path: str, delete_path: Optional[str] = None) -> Union[Datafile, bool]:
    """
    Convert a single Pubmed XML file to JSONL, pulling out any of the Pubmed entities and their upserts and or deletes.
    Used in parallelised transform section.

    :return: Datafile that's been transformed.
    """

    logging.info(f"Reading in file - {input_path}")

    try:
        data = load_datafile(input_path)
    except ValidationError:
        logging.info(f"Fields in XML are not valid against it's own DTD file - {input_path}")
        return False

    # Save upserts
    upserts = parse_articles(data)
    save_pubmed_jsonl(upsert_path, upserts)

    # Save deletes
    if delete_path is not None:
        deletes = parse_deletes(data)
        save_jsonl(delete_path, deletes)

    return input_path
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+        Return a glob pattern for the merged transform files.
+
+        :param entity_type: Type of the record, either either "upsert" or "delete".
+        :return: Retuns the path to the transformed file.
+        """
+        assert self.datafile_release is not None, "Datafile.transform_file_path: self.datafile_release is None"
+        return os.path.join(
+            self.datafile_release.transform_folder,
+            f"merged_{entity_type}_*-*_part_*.{self.file_type}",
+        )
```
jdddog, line 210: The file naming could be simpler, e.g. `upsert_merged_pubmed22n0003.jsonl.gz`

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+
+        :param entity_type: Type of the record, either "upsert" or "delete".
+        :return: Retuns the path to the transformed file.
+        """
+
+        assert self.datafile_release is not None, "Datafile.transform_file_path: self.datafile_release is None"
+        return os.path.join(
+            self.datafile_release.transform_folder,
+            f"{entity_type}_*.{self.file_type}",
+        )
```
jdddog, line 181: This method isn't used?

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1623 @@
...
+        # Remove the .gz compression if it's not a baseline file.
+        if self.baseline:
+            file_type = self.file_type
+        else:
+            file_type = self.file_type[:-3]
+
+        return os.path.join(
+            self.datafile_release.transform_folder,
+            f"{entity_type}_{self.filename[:-7]}.{file_type}",
+        )
```
jdddog, line 168: It might be easier to see what is going on by using @property and just hard coding different bits of the file paths.

```python

@property
def transform_baseline_path():
    return os.path.join(self.datafile_release.transform_folder, f"baseline_{self.filename[:-7]}.jsonl.gz")

@property
def transform_upsert_path():
    return os.path.join(self.datafile_release.transform_folder, f"upsert_{self.filename[:-7]}.jsonl")

@property
def merged_upsert_path():
    return os.path.join(self.datafile_release.transform_folder, f"upsert_merged_{self.filename[:-7]}.jsonl.gz")

@property
def delete_file_path():
    return os.path.join(self.datafile_release.transform_folder, f"delete.jsonl.gz")

```

jdddog (changes requested):
Hey Alex, thanks for the changes, they are looking good, I think we are nearly there. The main thing is that merging the upserts and deletes needs to take both the upserts and deletes into account (suggestions in comments).

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1559 @@
...
+                    record_keys[filename] = record_keys_per_file
+
+        assert len(datafiles) == len(
+            record_keys.keys()
+        ), f"Number of updatefiles does not match the number of keys in record_keys: {len(datafiles)} vs {len(record_keys.keys())}"
+
+        ti: TaskInstance = kwargs["ti"]
+        ti.xcom_push(key="record_keys", value=record_keys)
+
+    def merge_upserts_and_deletes(self, release: PubMedRelease, **kwargs):
```
jdddog, line 825: Whilst the output of this function is correct, it could be much more succinct:

```python
import dataclasses

@dataclasses.dataclass
class PMID:
    value: int
    version: int

    @property
    def pmid(self):
        return f"{self.value}-{self.version}"

    def __eq__(self, other):
        if not isinstance(other, self.__class__):
            return False
        return self.value == other.value and self.version == other.version

    def __hash__(self):
        return hash(self.pmid)


@dataclasses.dataclass
class PubmedUpdatefile:
    name: str
    upserts: List[PMID]
    deletes: List[PMID]

    @property
    def sequence(self):
        return int(re.search(r"(?<=pubmed)(\d+)(?=n)", self.name).group(1))

    @property
    def year(self):
        return int(re.search(r"(?<=n)(\d+)", self.name).group(1))

    def __eq__(self, other):
        if not isinstance(other, self.__class__):
            return False
        return self.name == other.name and set(self.upserts) == set(self.deletes)


def merge_upserts_and_deletes(updatefiles: List[PubmedUpdatefile]) -> Tuple[Dict[PMID, str], Set[PMID]]:
    upsert_index = dict()
    deletes = set()

    updatefiles.sort(key=lambda x: x.name)
    for file in updatefiles:
        # Apply upserts
        for upsert in file.upserts:
            # Check if item was previously deleted
            # and remove this delete if so
            if upsert in deletes:
                deletes.remove(upsert)

            # Update the file where the upsert comes from
            upsert_index[upsert] = file.name

        # Apply deletes
        for delete in file.deletes:
            # Delete any previous upserts
            if delete in upsert_index:
                del upsert_index[delete]

            # Add to delete set, as the record could still be in the main table
            deletes.add(delete)

    return upsert_index, deletes
```

The output of the function is the `upsert_index` dictionary, which you can directly use when iterating over the real updatefiles to determine which records to write out, only saving the upserts where the PMID key returns the same update file name.

You can also just write out the PMIDs in the deletes set straight after this function into one jsonl file, that doesn't need to be done in `save_pubmed_merged_records`, which can just be responsible for writing out the reconciled update files.

I would make the updatefile and PMID into a set of dataclasses and separate out the logic for merging the updates and deletes into its own function separate from the workflow so that it is easy to test and write a more in depth test case:
```python
def test_merge_upserts_and_deletes(self):
    changefiles = [
        PubmedUpdatefile(
            "pubmed10n2023",
            # Insert new records
            [PMID(10, 1), PMID(11, 1), PMID(12, 1), PMID(13, 1), PMID(14, 1), PMID(15, 1)],
            # Delete records from baseline
            [PMID(1, 1), PMID(2, 1)],
        ),
        PubmedUpdatefile(
            "pubmed11n2023",
            # Upsert over a previous record: PMID(10, 1)
            # Add a new version: PMID(12, 2), PMID(13, 2)
            # Add a new record: PMID(16, 1), PMID(17, 1)
            [PMID(10, 1), PMID(12, 2), PMID(13, 2), PMID(16, 1), PMID(17, 1)],
            # Delete a record from the previous day: PMID(11, 1)
            # Delete a record that was upserted on the same day: PMID(18, 1) n.b. this is removed from upserts
            [PMID(11, 1), PMID(18, 1)],
        ),
        PubmedUpdatefile(
            "pubmed12n2023",
            # Upsert over a previous record: PMID(10, 1), PMID(17, 1)
            # Add new records:
            [PMID(10, 1), PMID(17, 1), PMID(19, 1), PMID(20, 1), PMID(21, 1)],
            # Delete record added on previous day: PMID(13, 2)
            [PMID(13, 2)],
        ),
        PubmedUpdatefile(
            "pubmed13n2023",
            # Add a record deleted on a previous day: PMID(13, 2)
            # Add new versions:
            # Add new records:
            [PMID(13, 2), PMID(17, 1), PMID(19, 1), PMID(22, 1), PMID(23, 1)],
            # Delete records:
            [],
        ),
        PubmedUpdatefile(
            "pubmed14n2023",
            # Upsert new records:
            [PMID(24, 1), PMID(25, 1)],
            # Delete records that were previously upserted multiple times: PMID(17, 1)
            [PMID(17, 1)],
        ),
    ]
    expected_upserts = {
        PMID(12, 1): "pubmed10n2023",
        PMID(13, 1): "pubmed10n2023",
        PMID(14, 1): "pubmed10n2023",
        PMID(15, 1): "pubmed10n2023",
        PMID(12, 2): "pubmed11n2023",
        PMID(16, 1): "pubmed11n2023",
        PMID(10, 1): "pubmed12n2023",
        PMID(20, 1): "pubmed12n2023",
        PMID(21, 1): "pubmed12n2023",
        PMID(13, 2): "pubmed13n2023",
        PMID(19, 1): "pubmed13n2023",
        PMID(22, 1): "pubmed13n2023",
        PMID(23, 1): "pubmed13n2023",
        PMID(24, 1): "pubmed14n2023",
        PMID(25, 1): "pubmed14n2023",
    }
    expected_deletes = {PMID(1, 1), PMID(2, 1), PMID(11, 1), PMID(18, 1), PMID(17, 1)}

    # Check if we receive expected output
    actual_upserts, actual_deletes = merge_upserts_and_deletes(changefiles)
    self.assertEqual(expected_upserts, actual_upserts)
    self.assertEqual(expected_deletes, actual_deletes)
```

Then you can consolidate the tasks: `self.merge_upserts_and_deletes` and `self.save_merged_upserts_and_deletes` into one task.

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1559 @@
...
+        # Download and process the baseline files - skipped if no new baseline files are available.
+        self.add_task(self.download_baseline)
+        self.add_task(self.upload_downloaded_baseline)
+        self.add_task(self.transform_baseline)
+        self.add_task(self.upload_transformed_baseline)
+        self.add_task(self.bq_load_main_table)
+
+        # Create a backup of the main table before applying any changes.
+        # Only run if it is not the first run of the year.
+        self.add_task(self.create_snapshot)
```
jdddog, line 350: `self.add_task(self.create_snapshot)` should be before `download_baseline`.

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1559 @@
...
+        ### UPDATEFILES ###
+
+        # Download and pull out the upsert and delete records from the updatefiles.
+        self.add_task(self.download_updatefiles)
+        self.add_task(self.upload_downloaded_updatefiles)
+        self.add_task(self.transform_updatefiles)
+
+        # Determine which upserts and deletes are necessary to keep for this release.
+        self.add_task(self.merge_upserts_and_deletes)
+        self.add_task(self.save_merged_upserts_and_deletes)
```
jdddog, line 361: These two can be merged into one task, see comments below.

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1559 @@
...
+        Transform the *.xml.gz files downloaded from PubMed's FTP server into usable json-like files for BigQuery import.
+
+        This is a multithreaded and pulls the PubmedArticle records from the downloaded XML files.
+        """
+
+        # Get list of datafiles for the baseline to trasform.
+        datafiles = [datafile for datafile in release.datafile_list if not datafile.baseline]
+
+        # Object to store all of the upserts and delete keys that are present in each file.
+        record_keys = {}
```
jdddog, line 798: Given the comments below, you would want to output a set of `PubmedUpdatefile` instances, then convert them to JSON to send them over XCom, then parse them again in the next task. It adds a bit of boilerplate but it is easier to work with (and for writing tests) and makes it easier to follow what is going on.

jdddog (changes requested):
Hey Alex, I just have a few suggestions for simplifying the code for merging the updatefiles and writing out the deletes, once that is done and rebased it should be good to merge.

academic_observatory_workflows/workflows/pubmed_telescope.py
jdddog: So the final `merge_upserts_and_deletes` task would look something like:

```python
def merge_upserts_and_deletes(self, release: PubMedRelease, **kwargs):
    ti: TaskInstance = kwargs["ti"]
    updatefiles: List[PubmedUpdatefile] = parse_updatefiles(ti.xcom_pull(key="updatefiles"))

    # Merge records
    upsert_index, deletes = merge_upserts_and_deletes(updatefiles)

    # Save deletes
    save_jsonl_gz("/path/to/deletes.jsonl.gz", [delete.to_dict() for delete in deletes])

    # Save final upserts
    datafiles = [datafile for datafile in release.datafile_list if not datafile.baseline]
    for i, chunk in enumerate(get_chunks(input_list=datafiles, chunk_size=self.max_processes)):
        with ProcessPoolExecutor(max_workers=self.max_processes) as executor:
            logging.info(f"In chunk {i} and processing files: {chunk}")

            futures = {}
            for datafile in chunk:
                filename = os.path.basename(datafile.download_file_path)
                assert (
                    filename in upsert_index
                ), f"records for file: {filename} are not present in upsert_index. "

                future = executor.submit(
                    save_pubmed_merged_records,
                    upsert_index,
                    datafile.transform_upsert_file_path,
                    datafile.merged_upsert_file_path,
                )
                futures[future] = filename

            for future in as_completed(futures):
                filename = futures[future]
                logging.info(f"Finished writing out upserts and deletes for updatefile: {filename}")
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+    :param input_path: Where the original upsert records are stored.
+    :param output_path: Destination of where to write the merged upsert records.
+    :return filename: Name of the original updatefile, for logging purposes.
+    """
+
+    full_upsert_records = {str(record["MedlineCitation"]["PMID"]): record for record in load_jsonl(input_path)}
+    upserts_to_write = [full_upsert_records[record_key] for record_key in upserts]
+    save_pubmed_jsonl(output_path, upserts_to_write)
+
+    return filename
```
jdddog, line 1290:
```suggestion
def save_pubmed_merged_upserts(
    filename: str,
    upsert_index: Dict[PMID, str],
    input_path: str,
    output_path: str,
) -> str:
    """Used in parallel by save_merged_upserts_and_deletes to write out the merged upsert records using the custom
    Pubmed encoder.

    :param upserts: A list of upsert keys to pull out and write to file.
    :param input_path: Where the original upsert records are stored.
    :param output_path: Destination of where to write the merged upsert records.
    :return filename: Name of the original updatefile, for logging purposes.
    """

    data = []
    for record in yield_jsonl(input_path):
        pmid = PMID.from_dict(record["MedlineCitation"]["PMID"])
        if upsert_index[pmid] == filename:
            data.append(record)
    save_pubmed_jsonl(output_path, data)

    return filename
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+
+
+def merge_upserts_and_deletes(updatefiles: List[PubmedUpdatefile]) -> Tuple[Dict[str, List[str]], List[Dict]]:
```
jdddog, line 1293:
```suggestion
def merge_upserts_and_deletes(updatefiles: List[PubmedUpdatefile]) -> Tuple[Dict[PMID, str], Set[PMID]]:
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+    """Merge the Pubmed upserts and deletes and return them as a list of PMID value-Version pairs so that they can be
+    easily pulled from file.
+
+    :param updatefiles: List of PubmedUpdatefile objects to merge together.
+    :return: The merged upserts and deletes."""
+
+    upsert_index = dict()
+    deletes_set = set()
```
jdddog, line 1301:
```suggestion
    deletes = set()
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+    # Need to change back the upsert data structure from record[< PMID key obj >] = filename to record[filename] = List[string of keys]
+    upserts = {}
+    pmid: PMID
+    for pmid, filename in upsert_index.items():
+        if filename in upserts:
+            upserts[filename].append(pmid.to_str())
+        else:
+            upserts[filename] = [pmid.to_str()]
+
+    deletes = [delete.to_dict() for delete in deletes_set]
```
jdddog, line 1333:
```suggestion
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+    deletes_set = set()
+
+    updatefiles.sort(key=lambda x: x.name)
+    for file in updatefiles:
+        # Apply upserts
+        for upsert in file.upserts:
+            # Check if item was previously deleted
+            # and remove this delete if so
+            if upsert in deletes_set:
+                deletes_set.remove(upsert)
```
jdddog, line 1310:
```suggestion
                deletes.remove(upsert)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+            upsert_index[upsert] = file.name
+
+        # Apply deletes
+        for delete in file.deletes:
+            # Delete any previous upserts
+            if delete in upsert_index:
+                del upsert_index[delete]
+
+            # Add to delete set, as the record could still be in the main table
+            deletes_set.add(delete)
```
jdddog, line 1322:
```suggestion
            deletes.add(delete)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+    pmid: PMID
+    for pmid, filename in upsert_index.items():
+        if filename in upserts:
+            upserts[filename].append(pmid.to_str())
+        else:
+            upserts[filename] = [pmid.to_str()]
+
+    deletes = [delete.to_dict() for delete in deletes_set]
+
+    return upserts, deletes
```
jdddog, line 1335:
```suggestion
    return upsert_index, deletes
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+        """Merge the upserts and deletes for this release period."""
+
+        datafiles = [datafile for datafile in release.datafile_list if not datafile.baseline]
+
+        ti: TaskInstance = kwargs["ti"]
+        updatefile_list: list = ti.xcom_pull(key="updatefile_list")
+        updatefiles = [PubmedUpdatefile.from_dict(updatefile) for updatefile in updatefile_list]
+
+        # Merge records and return a list of what upserts to pull from the transformed updatefiles.
+        upserts, deletes = merge_upserts_and_deletes(updatefiles)
```
jdddog, line 905:
```suggestion
        upsert_index, deletes = merge_upserts_and_deletes(updatefiles)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+
+        # Save deletes
+        save_jsonl_gz(release.merged_delete_file_path, deletes)
```
jdddog, line 908:
```suggestion
        data = [delete.to_dict() for delete in deletes]
        save_jsonl_gz(release.merged_delete_file_path, data)
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+        # Save final upserts
+        for i, chunk in enumerate(get_chunks(input_list=datafiles, chunk_size=self.max_processes)):
+            with ProcessPoolExecutor(max_workers=self.max_processes) as executor:
+                logging.info(f"In chunk {i} and processing files: {chunk}")
+
+                futures = {}
+                datafile: Datafile
+                for datafile in chunk:
+                    filename = os.path.basename(datafile.download_file_path)
+                    assert filename in upserts, f"records for file: {filename} are not present in list of upserts. "
```
jdddog, line 919:
```suggestion
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+
+                    future = executor.submit(
+                        save_pubmed_merged_upserts,
+                        filename,
+                        upserts[filename],
```
jdddog, line 924:
```suggestion
                        upsert_index,
```

academic_observatory_workflows/workflows/pubmed_telescope.py
```diff
@@ -0,0 +1,1590 @@
...
+from airflow import AirflowException
+from airflow.operators.empty import EmptyOperator
+from airflow.models.taskinstance import TaskInstance
+
+from academic_observatory_workflows.config import schema_folder as default_schema_folder, Tag
+
+from observatory.platform.config import AirflowConns
+from observatory.platform.observatory_config import CloudWorkspace
+from observatory.api.client.model.dataset_release import DatasetRelease
+from observatory.platform.files import get_chunks, load_jsonl, save_jsonl_gz
```
jdddog, line 47:
```suggestion
from observatory.platform.files import get_chunks, save_jsonl_gz, yield_jsonl
```

jdddog (changes requested):
Thanks Alex, the changes look good. It is less work just to directly use the upsert_index and you are still able to parallelise the file writing, see the suggested changes for how to do that. The tests need a little tweaking to pass with these changes, although I think you can basically use the suggestion from the previous review for test_merge_upserts_and_deletes.

academic_observatory_workflows/workflows/pubmed_telescope.py
jdddog: Better to keep the dataclass versions around rather than the dict versions as these are easier to work with

jdddog (approved):
Great work Alex, it looks good to me!

PR #161, opened 2023-03-02
Fix/doi ror affiliation matcher errors

jdddog:
Fixing an issue where some affiliation strings will cause the ROR affiliation matcher to throw an HTTP 500 error. Catch the HTTP 500 errors and continue as if no ROR ids were matched for this entry, otherwise, re-raise the error as something else is likely wrong.


keegansmith21 (approved):
Looks good, thanks Jamie!

PR #162, opened 2023-03-06
Fix Crossref Metadata release date

jdddog:
Fix Crossref Metadata release date, the day of the month should be the last day of the month: e.g. `crossref_metadata20220907` changes to `crossref_metadata20220930`.

### DIFF
--- a/academic_observatory_workflows/workflows/tests/test_crossref_metadata_telescope.py
+++ b/academic_observatory_workflows/workflows/tests/test_crossref_metadata_telescope.py
@@ -18,15 +18,21 @@ import json
 import os
 from datetime import datetime
 from unittest.mock import patch
-import requests
 
 import httpretty
 import pendulum
+import requests
 from airflow.exceptions import AirflowException
 from airflow.models import Connection
 from airflow.utils.state import State
 from click.testing import CliRunner
 from natsort import natsorted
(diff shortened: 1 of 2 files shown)


academic_observatory_workflows/workflows/tests/test_crossref_metadata_telescope.py
```diff
@@ -18,15 +18,21 @@
...
 
 import httpretty
 import pendulum
+import requests
 from airflow.exceptions import AirflowException
 from airflow.models import Connection
 from airflow.utils.state import State
 from click.testing import CliRunner
 from natsort import natsorted
+from observatory.api.client.model.dataset import Dataset
```
keegansmith21, line 30: Cosmetic question - why are the imports ordered like this?

keegansmith21 (dismissed):
Changes look good Jamie, just a query about the import order. Merge when ready

academic_observatory_workflows/workflows/tests/test_crossref_metadata_telescope.py
jdddog: You're right it was odd. I'm not sure how they ended up like that! I use PyCharm to order them and removed unused imports, but not sure what it did.

PR #169, opened 2023-06-22
Orcid Telescope v2

keegansmith21:
# ORCID Telescope

This is a PR for the new and (hopefully) improved version of ORCID. Here's how it works

## Notes
- I took inspiration from the crossref events telescope - this will have a different logical flow for the first run through, then run the same for every subsequent run.
- The ORCID dataset is just a bucket of ~18 million files. This has been challenging to work around and many of the design choices had to account for this scale of file operation.
- The telescope has a lot of sanity checks. This is necessary because any missed files will only be picked up if they are altered in the source bucket in a subsequent run (there is no guarantee that they will ever be changed again!). So we don't want to let any records slip through the cracks.
- I also made an [s5cmd](https://github.com/peak/s5cmd) wrapper. It is extremely fast at downloading from a bucket and may be good to use for some other telescopes.

### Transfer
This step synchronises the source bucket with our local bucket. Any files that are missing or outdated will be written/overwritten with the source file. This uses the gsutil command and almost always fails on the first attempt as files are altered during the transfer. Usually the second attempt succeeds, but it will try up to 5 times (default) before giving up and raising an error.

### Create main table snapshot
For any run after the first, a snapshot of the main table is created. This is in case we break something in the run and need to revert the table

### Create manifests
For each ORCID batch (for which there are 1100), we create a manifest file that contains all of the blobs that we will process this run. We check which files need to be processed by comparing their update times to the _cutoff_ time. This time is the start of the epoch for the initial run. For subsequent runs it is set to the previous run's most recent update time (which is stored in the release information in the API). We also create a master manifest file that joins each batch manifest.

### Download
The download step utilises s5cmd to download as quickly as possible. I found this to be at least 10 times faster than gsutil's parallelised download method. 

We execute one download command per ORCID batch by reading that batch's manifest and piping the blob names to the s5cmd command. After each download, the telescope checks that the expected files have indeed been downloaded. 

### Transform
The transform step  converts the XML format into a dictionary, fixes up the keys and saves to JSON. Each batch of records is converted into two files. The first is all of the records in a gzipped jsonl file called the _upsert_ records. The second file contains all of the _delete_ records - which are records that need to be removed. 

Throughout the transform process, checks are made to ensure that the number of records processed per batch is equal to the number of files.

### BQ Load main table
This is only run in the initial run. All of the transformed upsert records are loaded into a single table - the main table. There is no point doing anything with the _delete_ records since they are only used as an indication of which records to remove.

### BQ Load upsert table | BQ Load delete table
For each run after the first, the upsert|delete records are loaded into their own bigquery table

### BQ upsert records
All records in the upsert table are upserted into the main table.

### BQ delete records
All records in the main table that have an ID in the delete table are removed from the main table.

### Add new dataset release
One thing of note here is that the master manifest file is scanned before the release is created. This is to check for the most recent update time. This is used in the next create_manifests step as the _cutoff_ time.


keegansmith21:
Tests are just waiting on [#635](https://github.com/The-Academic-Observatory/observatory-platform/pull/635)

academic_observatory_workflows/s5cmd.py
```diff
@@ -0,0 +1,172 @@
...
+        ]
+        cfg = [i for i in cfg if i]  # Remove empty strings
+        return " ".join(cfg)
+
+
+class S5Cmd:
+    def __init__(
+        self,
+        access_credentials: Tuple[str, str],
+        out_stream: str = PIPE,
```
jdddog, line 46: The `out_stream` variable should have a type of `Union[None, int, TextIO]` which matches the `Popen` `stdout` and `stderr` types:

```python
out_stream: Union[None, int, TextIO] = PIPE
```

You can import `TextIO` from `from typing import TextIO`.

academic_observatory_workflows/s5cmd.py
```diff
@@ -0,0 +1,172 @@
...
+        logging.info(f"s5cmd blob download command example: {blob_cmds[0]}")
+
+        # Initialise credentials and execute
+        with self._bucket_credentials() as credentials:
+            cmd += f" --credentials-file {credentials} run"
+            proc = Popen(shlex.split(cmd), stdout=self.out_stream, stderr=self.out_stream, stdin=PIPE)
+            stdout, stderr = proc.communicate(input=blob_stdin.encode())
+        returncode = proc.wait()
+        if returncode > 0:
+            logging.warn(f"s5cmd cp failed with return code {returncode}: {stderr}")
```
jdddog, line 135: `logging.warn` is deprecated and should be `logging.warning`.

requirements.sh
```diff
@@ -26,3 +26,10 @@ apt-get install pigz unzip -y
 echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | \
 tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
 apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - && apt-get update -y && apt-get install google-cloud-sdk -y
+
+# ORCID install s5cmd
+curl -LO https://github.com/peak/s5cmd/releases/download/v2.1.0/s5cmd_2.1.0_linux_amd64.deb && \
+dpkg -i s5cmd_2.1.0_linux_amd64.deb
+#chmod +x s5cmd_2.1.0_linux_amd64.deb | \
+#apt install ./s5cmd_2.1.0_linux_amd64.deb -y 
+
```
jdddog, line 35: Can these lines be deleted?

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,756 @@
...
+        bq_create_dataset(
+            project_id=self.cloud_workspace.project_id,
+            dataset_id=self.bq_dataset_id,
+            location=self.cloud_workspace.data_location,
+            description=self.dataset_description,
+        )
+
+    def transfer_orcid(self, release: OrcidRelease, **kwargs):
+        """Sync files from AWS bucket to Google Cloud bucket."""
+        for i in range(self.transfer_attempts):
```
jdddog, line 366: `success` could be undefined if `self.transfer_attempts` was set to zero or negative, so better to set `success=False` before the loop.

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,756 @@
...
+        if total_upsert_records + total_delete_records != len(release.downloaded_records):
+            raise ValueError(
+                f"Expected {len(release.downloaded_records)} total records processed but got {total_upsert_records + total_delete_records}"
+            )
+
+    def upload_transformed(self, release: OrcidRelease, **kwargs):
+        """Uploads the upsert and delete files to the transform bucket."""
+        success = gcs_upload_files(bucket_name=self.cloud_workspace.transform_bucket, file_paths=release.upsert_files)
+        set_task_state(success, self.upload_transformed.__name__, release)
+        success = gcs_upload_files(bucket_name=self.cloud_workspace.transform_bucket, file_paths=release.delete_files)
```
jdddog, line 524: Perhaps simpler to just do:

```python
file_paths = release.upsert_files + release.delete_files
success = gcs_upload_files(bucket_name=self.cloud_workspace.transform_bucket, file_paths=file_paths)
set_task_state(success, self.upload_transformed.__name__, release)
```

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,756 @@
...
+        cleanup(dag_id=self.dag_id, execution_date=kwargs["logical_date"], workflow_folder=release.workflow_folder)
+
+
+def create_orcid_batch_manifest(
+    orcid_batch: OrcidBatch, reference_date: pendulum.DateTime, bucket: str, bucket_prefix: str = None
+) -> None:
+    """Create a manifest (.csv) for each orcid batch containing blob names and modification dates of all the
+    modified files in the bucket's orcid directory. Only blobs modified after the reference date are included.
+
+    :param orcid_directory: The OrcidBatch instance for this orcid directory
```
jdddog, line 651:
```suggestion
    :param orcid_batch: The OrcidBatch instance for this orcid directory
```

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,756 @@
...
+
+    with open(orcid_batch.manifest_file, "w", newline="") as csvfile:
+        writer = csv.DictWriter(csvfile, fieldnames=MANIFEST_HEADER)
+        writer.writeheader()
+        writer.writerows(manifest)
+
+    logging.info(f"Manifest saved to {orcid_batch.manifest_file}")
+
+
+def latest_modified_record_date(manifest_file_path) -> pendulum.DateTime:
```
jdddog, line 680: Could set a type on `manifest_file_path`:

```python
from os import PathLike

def latest_modified_record_date(manifest_file_path: Union[str, PathLike]) -> pendulum.DateTime:
```

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,756 @@
...
+
+    :return: A list of all the possible ORCID directories
+    """
+    n_1_2 = [str(i) for i in range(10)]  # 0, 1, 2, 3, 4, 5, 6, 7, 8, 9
+    n_3 = n_1_2.copy() + ["X"]  # 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, X
+    combinations = list(itertools.product(n_1_2, n_1_2, n_3))  # Creates the 000 to 99X directory structure
+    return ["".join(i) for i in combinations]
+
+
+def transform_orcid_record(record_path: str) -> Union(Dict, str):
```
jdddog, line 703: Should use square brackets with Union:
```suggestion
def transform_orcid_record(record_path: str) -> Union[Dict, str]:
```

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,756 @@
...
+    def missing_records(self) -> List[str]:
+        """List of missing ORCID records on disk for this ORCID directory."""
+        return list(set(self.expected_records) - set(self.existing_records))
+
+    @property
+    def blob_uris(self) -> List[str]:
+        """List of blob URIs from the manifest this ORCID directory."""
+        with open(self.manifest_file, "r") as f:
+            reader = csv.DictReader(f)
+            return [gcs_blob_uri(bucket_name=row["bucket_name"], blob_name=row["blob_name"]) for row in reader]
```
jdddog, line 125: It may be beneficial to read the `manifest_file` in the constructor, so that it doesn't have to be read in each time a property is called.

Also, for a given `manifest_file`, do the results returned by each property ever change while that instance is alive? If not, you could use of `cached_property` from `from functools import cached_property` as this will cache the results if the property is called more than one time.

See: https://docs.python.org/3/library/functools.html#functools.cached_property

jdddog (changes requested):
Hey Keegan, the workflow looks very good. I've mostly left some small suggestions related to the type hints and the OrcidBatch properties.

The `load_and_parse_json` function changes breaks the tests for the DOI Workflow, Crossref Events, OpenAlex and Unpaywall, so it would be good if the changes to `load_and_parse_json` were done in a way that doesn't break those other tests, see here for the output: https://github.com/The-Academic-Observatory/observatory-platform/pull/638#issuecomment-1710971612

academic_observatory_workflows/workflows/orcid_telescope.py
keegansmith21: I didn't know this cached_property existed, this is really good! 

The expected_records and blob_uris properties aren't expected to ever change for any given instance so they can be cached. However the other two should be calculated each time they're called. 

The manifest file doesn't necessarily exist on construction, so I'd rather not put it in the constructor as it'd introduce some ugly logic.

academic_observatory_workflows/workflows/orcid_telescope.py
keegansmith21: Yep, didn't think of this

academic_observatory_workflows/workflows/orcid_telescope.py
keegansmith21: Yeah that's nicer

academic_observatory_workflows/workflows/orcid_telescope.py
keegansmith21: Sure, is this just a convention thing?

requirements.sh
keegansmith21: They sure can!

academic_observatory_workflows/workflows/orcid_telescope.py
jdddog: Yeah it is just how type hints are written, e.g. https://docs.python.org/3/library/typing.html#typing.Union

academic_observatory_workflows/workflows/orcid_telescope.py
jdddog: Yeah, I haven't used it before either but should. OK fair enough.

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,758 @@
...
+        with open(self.manifest_file, "r") as f:
+            reader = csv.DictReader(f)
+            return [os.path.basename(row["blob_name"]) for row in reader]
+
+    @cached_property
+    def blob_uris(self) -> List[str]:
+        """List of blob URIs from the manifest this ORCID directory."""
+        with open(self.manifest_file, "r") as f:
+            reader = csv.DictReader(f)
+            return [gcs_blob_uri(bucket_name=row["bucket_name"], blob_name=row["blob_name"]) for row in reader]
```
jdddog, line 127:
```suggestion
        return [gcs_blob_uri(bucket_name=row["bucket_name"], blob_name=row["blob_name"]) for row in self._manifest_data]
```

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,758 @@
...
+    def missing_records(self) -> List[str]:
+        """List of missing ORCID records on disk for this ORCID directory."""
+        return list(set(self.expected_records) - set(self.existing_records))
+
+    @cached_property
+    def expected_records(self) -> List[str]:
+        """List of expected ORCID records for this ORCID directory. Derived from the manifest file"""
+        with open(self.manifest_file, "r") as f:
+            reader = csv.DictReader(f)
+            return [os.path.basename(row["blob_name"]) for row in reader]
```
jdddog, line 120:
```suggestion
        return [os.path.basename(row["blob_name"]) for row in self._manifest_data]
```

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,758 @@
...
+
+        if not os.path.exists(self.download_dir):
+            raise NotADirectoryError(f"Directory {self.download_dir} does not exist.")
+        if not os.path.exists(self.transform_dir):
+            raise NotADirectoryError(f"Directory {self.transform_dir} does not exist.")
+        if not re.match(BATCH_REGEX, self.batch_str):
+            raise ValueError(f"Batch string {self.batch_str} is not valid.")
+
+        os.makedirs(self.download_batch_dir, exist_ok=True)
+
```
jdddog, line 104:
```suggestion

    @cached_property
    def _manifest_data(self) -> List[Dict]:
        with open(self.manifest_file, "r") as f:
            return list(csv.DictReader(f))

```

jdddog (changes requested):
Hey Keegan, thanks for making the changes, that looks good. That's a good point about the manifest file not existing on construction - I think that would be a good use of a cached property though - see the suggestion below.

academic_observatory_workflows/workflows/orcid_telescope.py
```diff
@@ -0,0 +1,761 @@
...
+            f"Expected ORCID {expected_orcid} does not match ORCID in record {orcid_record['common:orcid-identifier']['common:path']}"
+        )
+
+    # Transform the keys of the dictionary so they are valid BigQuery fields
+    orcid_record = change_keys(orcid_record)
+
+    return orcid_record
+
+
+def change_keys(obj):
```
alexmassen-hane, line 746: This type of function already exists in observatory platform. Consider using this instead:

https://github.com/The-Academic-Observatory/observatory-platform/blob/b94c968879b4bef802c0012a294b616637c68be0/observatory-platform/observatory/platform/files.py#L531C15-L531C15

alexmassen-hane (changes requested):
Thanks Keegan, looks good. Just a small suggestion for using a function that we already have in the platform for changing keys.

academic_observatory_workflows/workflows/orcid_telescope.py
keegansmith21: Thanks Alex. Yeah I think I initially used this function, but was experimenting with speeding up the transform step. I concluded that the way I implemented wasn't significantly faster, so I may as well change it back to use the common function. :+1:

jdddog (approved):
Thanks Keegan, looks good to me.

PR #170, opened 2023-06-26
Fix write disposition

jdddog:
OpenCitations shouldn't be using append and the main table of Crossref Events can use the default write empy.


keegansmith21 (approved):
Changes look good. Thanks Jamie!

PR #171, opened 2023-07-03
Refactor DOI table

jdddog:
Refactors the DOI table:
* Moved the real Unpaywall data into its own struct.
* Moved COKI generated data into the coki struct.
* Add Black OA to coki.oa.colors based on whether DOI is in SciHub.
* Removed MAG from DOI table.
* Removed Elastic export tables.
* Updated test data for oa web workflow.
* Made Unpaywall test data not cover all outputs from Crossref Metadata, which is how the datasets work in reality.


academic_observatory_workflows/model.py
```diff
@@ -805,9 +817,16 @@ def make_papers(
             publisher_license=license_,
             publisher_is_free_to_read=publisher_is_free_to_read_,
             repositories=paper_repos,
+            in_scihub=bool(random.getrandbits(1)),
```
keegansmith21, line 820: What's the purpose of this randomness?

keegansmith21 (approved):
Looks good Jamie! I just had a question about a line, but feel free to merge when ready

academic_observatory_workflows/model.py
jdddog: Ah, just to make it so that 50% of the generated test papers are found in SciHub and so that 50% aren't

PR #172, opened 2023-07-25
Fix/Add s3:// prefix to OpenAlex AWS blob URL if not present

alexmassen-hane:
Sometimes URLs for AWS blobs do not have the s3:// prefix, but it is required in later parts of the workflow for transferring files from the AWS bucket to GCS.

This PR adds the prefix in the ManifestEntry class if the URL does not start with it.
Assert is added in the ManifestEntry test to make sure that it is a valid AWS s3 bucket URL.


jdddog (approved):
Thanks for fixing the issue with OpenAlex, Alex, it looks good!

PR #173, opened 2023-07-26
Fix/OpenAlex "abstract_inverted_index" field is sometimes a string, not a dictionary

alexmassen-hane:
For the recent 'works' output for the OpenAlex data dump, the "abstract_inverted_index" field can sometimes be json string dump. This PR adds the ability to decode the string dump and pull out the keys and values as needed. The resulting data should still be in the same format as before and it will not require a re-run of the entire workflow. 

Please see the post on the Google forum for OpenAlex:
https://groups.google.com/g/openalex-users/c/TsVgxd_GEuw

### DIFF
--- a/academic_observatory_workflows/workflows/openalex_telescope.py
+++ b/academic_observatory_workflows/workflows/openalex_telescope.py
@@ -1016,12 +1016,30 @@ def transform_object(obj: dict):
 
     field = "abstract_inverted_index"
     if field in obj:
-        if not isinstance(obj.get(field), dict):
+        if not isinstance(obj.get(field), (dict, str)):
             return
-        keys = list(obj[field].keys())
-        values = [str(value)[1:-1] for value in obj[field].values()]
+        else:
+            # If data is held in a string dump, load json string again.
+            if isinstance(obj.get(field), str):
+                obj_part = json.loads(obj[field])
+                field2 = "InvertedIndex"
+                if isinstance(obj_part.get(field2), dict):
+                    keys = list(obj_part[field2].keys())
+                    values = [str(value)[1:-1] for value in obj_part[field2].values()]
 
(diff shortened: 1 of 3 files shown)


academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -1016,12 +1016,30 @@ def transform_object(obj: dict):
...
+
+                    obj[field] = {"keys": keys, "values": values}
+                else:
+                    raise TypeError(f"obj_part['InvertedIndex'] is not a dictionary: {obj_part}")
+            else:
+                keys = list(obj[field].keys())
+                values = [str(value)[1:-1] for value in obj[field].values()]
 
-        obj[field] = {"keys": keys, "values": values}
+                obj[field] = {"keys": keys, "values": values}
```
jdddog, line 1042: I don't think we need to check if the index length matches, as it is a problem for OpenAlex to fix if that is the case. It could be a data quality check that we do in the future (if we re-ran the workflow and updated the field).

You could simplify the code a bit by putting the parsing into a function, like this:

```python
    field = "abstract_inverted_index"
    if field in obj:
        def parse_abstract(dict_: dict):
            keys_ = list(dict_.keys())
            values_ = [str(value_)[1:-1] for value_ in dict_.values()]
            return {"keys": keys_, "values": values_}
        if isinstance(obj.get(field), str):
            data = json.loads(obj[field])
            obj[field] = parse_abstract(data["InvertedIndex"])
        elif isinstance(obj.get(field), dict):
            obj[field] = parse_abstract(obj[field])
        else:
            return
```

jdddog (changes requested):
Thanks for fixing this Alex, I just have a small suggestion, other than that it looks good.

jdddog (approved):
Thanks Alex, it looks good.

PR #174, opened 2023-07-26
INF-638: Data Quality checks for production tables

alexmassen-hane:
# Data Quality Check Workflow

This PR introduces basic data quality checks for data produced by telescopes in the Academic Observatory Workflows. For each dataset defined in the workflow, it will create a metadata table gathered for all of the tables in that dataset. The metadata produced for each table is held within a 'data quality check' (DQC) record and is stored as a many-to-many relationship. 

Each DQC record holds the following metadata about the each of the tables in the dataset:
- table_name (str): Friendly table name.
- full_table_id (str): Fully qualified id of the table in Google Bigquery f"{project_id}.{dataset_id}.{table_id}"
- hash_id (str): Based off of the full_table_id, number of fields, number of rows and number of bytes of stored in the table.
- sharded (bool): If the table is sharded or not. 
- date_shard (date): Date from the table shard (null if it is not a sharded table).
- date_created (datetime): Date of when the table was created. 
- expires (bool): If the table is set to expire or not. 
- date_expires (datetime): Date of when the table expires, null if it not to expire.
- date_last_modified (datetime): Date of when the table was modified.
- date_checked (datetime): Date of when the table was checked by this workflow.
- Size_Gb (float): The size of the table in Bigquery in gigabytes. 
- primary_key ( List[str] ): The fields/keys uses to identify records of a table. 
- num_rows (int): Number of rows in the table.
- num_distinct_records (int): Number of records with distinct primary keys.
- num_null_records (int): Number of records that have no value for "primary_key".
- num_duplicates (int): Number of records that have duplicate values under the "primary_key".
- num_fields (int): Total number of fields in the table. This is see how the table grows/modified overtime. 

## Notes
- The last few fields of the DQC record (num_distinct_records,  num_null_records, num_duplicates) are determined by using small SQL quires that have been optimised to use as little quota as possible. 
- This workflow was designed to be as lightweight as possible. IT=t does not require a connection to the Observatory API and it quires the metadata tables it produces to determine if it needs to perform the check a on table again. 
- The table hash that is used to determine if a table has already been checked uses a combination of the name of the table, number of bytes stored, number of rows and the total number of fields, which is a cheap and quick way of roughly determining a md5hash for the entire table. However, this assumes that changes to the data in the table is a significant number of bytes and rows different. Typically, upserts or deletes operations for tables in the Academic Observatory change tables on the order of megabytes and 1000s of rows so the changes of the table hashes being the same if negligible. 
- This workflow includes the ability to check over datasets that are not just produced by AO Workflows, for example OpenAIRE has been manually added to the project and will be checked just the same as other tables.

### Dag Sensors
Similarly to the DOI Workflow, it will wait until all of the dags defined have finished before starting the next run of the workflow.

### Create Dataset
This step creates the dataset that will hold the metadata tables created by this workflow.

### Perform Data Quality Check
Each dataset checked over by this workflow is run as a parallel task. The structure of the workflow will look something like the following, depending on the number of datasets included in the workflow:

image


jdddog (changes requested):
Hey Alex, the workflow looks well structured, well documented and the type hints look pretty good.

At a high level there are a couple of things that could be tweaked:

* It would be easier to write / query the data if we output the data into a single table, e.g. data_quality.
* The DOI workflow has issues with the sensor looking at multiple DAG runs, where different DAGs run on different schedules.
  * There are actually a number of different ways that this could be designed. I would prefer to use the release API, but that would need a little bit of a modification and it won't support datasets that are manually uploaded.
  * So instead, we could make it so that there is a data quality workflow instance for each workflow / dataset:
    * Configure the sensor (if required) and DAG settings so that they match up with the workflow that generates the data. The sensor, dataset and tables settings can be configured via the kwargs in the workflow config. Using the Airflow sensor rather than ours would be better.
    * For manually uploaded datasets, just set no sensor, and have it run on a schedule.
* It would be good to define the table information in the workflow config file. I'll send a version of this to you directly.
* What happens if we delete a DAG / it's data and re-run it? E.g. OpenAlex.
* Minor point: you are using both logging.info() and print(). It would be better to use the logging module as it provides better control over log levels and outputs.

alexmassen-hane:
I have made the requested changes, but the unit tests will fail as it relies on the following PR:
https://github.com/The-Academic-Observatory/observatory-platform/pull/646

The DQC workflows are now defined using the config.yaml for the platform with the following structure:
```
  - dag_id: "dqc_openalex"
    name: "DQC Workflow - OpenAlex"
    class_name: "academic_observatory_workflows.workflows.dqc_workflow.DQCWorkflow"
    cloud_workspace: *academic_observatory
    kwargs:
      schedule: "@weekly"
      sensor_dag_ids: ["openalex"]
      datasets:
        openalex:
          tables:
            - table_id: authors
              is_sharded: false
              fields: [id]
            - table_id: works
              is_sharded: false
              fields: [id]
```
and will produce a workflow like the following:
image

The workflow can handle multiple dag_sensors_ids and multiple datasets if necessary. If there are multiple tables listed in the config file, it will process each table in parallel. The dag_sensor_ids field can be ignored for when there are manually uploaded datasets into the project, such as OpenAIRE and will run on the specified schedule. Since this can now spawn a number of different workflows, I have also added a data_quality_check tag so we can easily filter them in the Airflow GUI.

academic_observatory_workflows/workflows/dqc_workflow.py
```diff
@@ -0,0 +1,456 @@
...
+        self.bq_dataset_id = bq_dataset_id
+        self.bq_table_id = bq_table_id
+        self.bq_dataset_description = bq_dataset_description
+        self.bq_table_description = bq_table_description
+        self.data_location = cloud_workspace.data_location
+        self.schema_path = schema_path
+        self.datasets = datasets
+
+        # Full table id for the DQC records.
+        self.dqc_full_table_id = f"{self.project_id}.{self.bq_dataset_id}.{bq_table_id}"
```
jdddog, line 152:
```suggestion
        self.dqc_full_table_id = make_bq_table_id(self.project_id, self.bq_dataset_id, bq_table_id)
```

academic_observatory_workflows/workflows/dqc_workflow.py
```diff
@@ -0,0 +1,456 @@
...
+
+        with self.dag:
+            # Create a group of sensors for the workflow.
+            task_sensor_group = []
+            if self.sensor_dag_ids:
+                with TaskGroup(group_id="dag_sensors") as tg_sensors:
+                    task_sensors = []
+
+                    for ext_dag_id in self.sensor_dag_ids:
+                        sensor = DagRunSensor(
```
jdddog, line 172: It would be good to change this to use the Airflow `ExternalTaskSensor` which should be more reliable than the DagRunSensor. Some of its parameters may need configuring so that it waits for the given DAG.

https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/sensors/external_task/index.html

academic_observatory_workflows/workflows/dqc_workflow.py
```diff
@@ -0,0 +1,456 @@
...
+        """Create a dataset for all of the DQC tables for the workflows."""
+
+        success = bq_create_dataset(
+            project_id=self.project_id,
+            dataset_id=self.bq_dataset_id,
+            location=self.data_location,
+            description=self.bq_dataset_description,
+        )
+
+        set_task_state(success, self.create_dataset.__name__, release)
```
jdddog, line 240: This function doesn't return success, I think it returns the dataset and may throw an exception if it can't create the dataset? So you can probably remove success and set task state.

academic_observatory_workflows/workflows/dqc_workflow.py
```diff
@@ -0,0 +1,456 @@
...
+    num_all_fields = len(bq_select_columns(table_id=full_table_id))
+
+    return dict(
+        table_id=table_id,
+        dataset_id=dataset_id,
+        project_id=project_id,
+        full_table_id=full_table_id,
+        hash_id=hash_id,
+        is_sharded=is_sharded,
+        shard_date=shard_date.strftime("%Y-%m-%d") if shard_date is not None else shard_date,
```
jdddog, line 353:
```suggestion
        shard_date=shard_date.to_date_string() if shard_date is not None else shard_date,
```

academic_observatory_workflows/workflows/dqc_workflow.py
```diff
@@ -0,0 +1,456 @@
...
+                        table_tasks.append(task)
+
+                    # Add the list of table_tasks to tg_dataset by this line
+                    (table_tasks)
+
+                    task_datasets_group.append(tg_dataset)
+
+            # Link all tasks and task groups together.
+            (task_sensor_group >> task_check_dependencies >> task_create_dataset >> task_datasets_group)
+
```
jdddog, line 217: It would be a bit cleaner to separate the parsing of the "datasets" and "tables" from creating the DAG. You could make a function that converts the datasets / tables passed to the workflow into Table dataclass objects etc.

academic_observatory_workflows/workflows/dqc_workflow.py
```diff
@@ -0,0 +1,456 @@
...
+                            duration=timedelta(days=7),
+                            poke_interval=int(timedelta(hours=1).total_seconds()),
+                            timeout=int(timedelta(days=3).total_seconds()),
+                        )
+                        print(sensor)
+                        self.add_operator(sensor)
+                        task_sensors.append(sensor)
+
+                    # Add all of the sesnors to tg_sensors with this line.
+                    (task_sensors)
```
jdddog, line 185: I don't think this statement does anything?

academic_observatory_workflows/workflows/dqc_workflow.py
```diff
@@ -0,0 +1,456 @@
...
+                    for table in tables:
+                        task = self.make_python_operator(
+                            self.perform_data_quality_check,
+                            task_id=f"{table.table_id}",
+                            op_kwargs={f"task_id": f"{table.table_id}", "table": table},
+                        )
+                        table_tasks.append(task)
+
+                    # Add the list of table_tasks to tg_dataset by this line
+                    (table_tasks)
```
jdddog, line 211: I don't think this statement does anything?

jdddog (changes requested):
Hey Alex, thanks for the changes, it is looking good. I've just left feedback about a few smaller things. Also, I think it would be be easier to know what the workflow does by calling it the DataQualityWorkflow rather than DQCWorkflow.

Could you also add the configuration for the workflows to the workflows-config project?

academic_observatory_workflows/workflows/tests/test_data_quality_workflow.py
```diff
@@ -0,0 +1,544 @@
...
+        env = ObservatoryEnvironment(self.project_id, self.data_location)
+
+        # Where the metadata generated for this workflow is going to be stored.
+        dqc_dataset_id = env.add_dataset(prefix="data_quality_check")
+        fake_dataset_id = env.add_dataset()
+        settings_dataset_id = env.add_dataset(prefix="settings")
+
+        # Borrowing from the DOI Workflow to create fake data for these tests
+        repositories = TestDoiWorkflow().repositories
+        institutions = TestDoiWorkflow().institutions
```
jdddog, line 135: It is better not to import TestDoiWorkflow like this, because at least with PyCharm, it will then also run the TestDoiWorkflow tests as well.

It would be better to load a dataset that you can use to check that the data quality workflow is producing correct results, you could even just write it by hand in JSON files, since you really just need an ID field for each table. You want to test that the workflow can operate on sharded and non sharded tables, single fields, and mutliple nested fields (like the pubmed example). Then also test that the output is correct - so have some duplicate IDs, null ids etc, which might not be present in the fake DOI Workflow data.

academic_observatory_workflows/workflows/data_quality_workflow.py
```diff
@@ -0,0 +1,437 @@
...
+        )
+        for table in tables
+    ]
+
+
+class DataQualityWorkflow(Workflow):
+    def __init__(
+        self,
+        *,
+        dag_id: str,
```
jdddog, line 97: If we run this on all of the past BigQuery sharded tables, it will pull many of them out of long term storage, which would suddenly increase our BigQuery storage costs if we run it on everything. So we should add a parameter where you can specify the maximum number of past shards to analyse.

jdddog (changes requested):
Thanks for the changes Alex. It looks nearly ready to go, we just need to add a parameter so that we don't analyse all of the past shards, and a change to the tests. Then I we can merge it and start with a small number of datasets / tables to check.

alexmassen-hane:
I have added the "num_shards" for the input on the config for the workflow so that we can limit the number of shards processed to reduce Bigquery costs. I have also renamed the "fields" parameter to "primary_key" as we are going to expand the data quality checks to a select number of fields (or possibly all fields) in the near future.

PR #179, opened 2023-08-25
Ignore ROR relationships where the ROR IDs don't exist

jdddog:
Ignore ROR relationships where the ROR IDs don't exist

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -24,6 +24,8 @@ jobs:
 
       - name: Install dependencies
         run: |
+          sudo apt-get install pigz unzip -y
```
jdddog, line 27:
```suggestion
          sudo apt-get install pigz -y
```

PR #181, opened 2023-08-30
Fix/Pubmed Telescope: Merge upserts and deletes

alexmassen-hane:
A proposed fix for the multiprocessing step in the Pubmed telescope for writing out the merged upserts and deletes. 
The function `save_merged_upserts_and_deletes` now returns the path to the file that it writes to. The task will now only log the path if the spawned process in the process pool has completed, using `future.result()`. 

Also added logging to the `merge_upserts_and_deletes` function so that we know it is processing each updatefile properly.


jdddog (approved):
Thanks for fixing this Alex

PR #183, opened 2023-09-07
Fix/OpenAlex: abstract_inverted_index field now a JSON structure

alexmassen-hane:
A proposed fix for the new structure of the abstract_inverted_index field as it is now a proper JSON object. Also added to the unit tests to make sure it transforms to the desired format.


jdddog (approved):
Thanks for fixing this Alex!

PR #185, opened 2023-09-14
INF-652:Add new fields to OpenAlex

alexmassen-hane:
This PR is to preemptively add new fields to the OpenAlex tables before the release of the September snapshot.

As per this thread [https://groups.google.com/g/openalex-users/c/Gs5rIwQexIo](url), the following fields will be added to tables:

works:
- countries_distinct_count (integer)
- institutions_distinct_count (integer)
- locations.Is_accepted (bool)
- locations.is_published (bool)
- authorships.raw_author_name (string)
- sustainable_development_goals (record of; id (string), display_name (string), score (float) )

institutions:
- lineage ( list of strings )


jdddog (approved):
Thanks Alex, the schema updates look good to me.

PR #187, opened 2023-09-25
OpenAlex: use taskgroups and process each table as separate tasks

jdddog:
This PR does the following:
* Updates the OpenAlex schemas for the September 2023 release.
* Refactors the OpenAlex workflow to use Airflow Task Groups, processing each OpenAlex entity in separate task groups, so that a failure in processing one entity means that the operations for the other entities don't need to be repeated or delayed. See image below for an example. 
* Remove code for only transforming some entities to simplify the workflow. The works table is the largest and has to be transformed anyway.

image

Tests will fail as it depends on: https://github.com/The-Academic-Observatory/observatory-platform/pull/642


academic_observatory_workflows/database/schema/openalex/institutions.json
```diff
@@ -290,49 +268,49 @@
...
-        "name": "issn",
+        "name": "host_organization_lineage",
         "type": "STRING",
-        "mode": "REPEATED"
+        "mode": "REPEATED",
+        "description": "The host organisations lineage."
       },
       {
-        "name": "display_name",
+        "name": "host_organization_name",
```
alexmassen-hane, line 289: I think this field is "host_organization_lineage_names" and is given as a list of strings.

academic_observatory_workflows/database/schema/openalex/works.json
```diff
@@ -167,7 +155,7 @@
...
+    "description": "Object: Information about the paid APC (article processing charge) for this work. You can find the listed APC price (when we know it) for a given work using apc_list. However, authors don\u2019t always pay the listed price; often they get a discounted price from publishers. So it\u2019s useful to know the APC actually paid by authors, as distinct from the list price. This is our effort to provide this. Our best source for the actually paid price is the OpenAPC project. Where available, we use that data, and so apc_paid.provenance is openapc. Where OpenAPC data is unavailable (and unfortunately this is common) we make our best guess by assuming the author paid the APC list price, and apc_paid.provenance will be set to wherever we got the list price from."
   },
   {
     "name": "authorships",
```
alexmassen-hane, line 161: The institution object under authorships is missing the new lineage field.

alexmassen-hane (dismissed):
Looks good Jamie. A field needs to be renamed and there's one pesky one of the new lineage fields missing.

academic_observatory_workflows/database/schema/openalex/institutions.json
jdddog: This one is still a valid field: https://docs.openalex.org/api-entities/sources/source-object#host_organization_name

academic_observatory_workflows/database/schema/openalex/works.json
jdddog: Thanks, I've added this one.

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -356,53 +363,70 @@ def __init__(
...
-
-        # The last task that the next DAG run's ExternalTaskSensor waits for.
-        self.add_operator(
-            EmptyOperator(
-                task_id=external_task_id,
+            task_check_dependencies = PythonOperator(python_callable=self.check_dependencies, task_id="check_dependencies")
+            task_fetch_releases = ShortCircuitOperator(python_callable=self.fetch_releases, task_id="fetch_releases") # If there are no changes then skip all below tasks
+            task_create_datasets = PythonOperator(python_callable=self.create_datasets, task_id="create_datasets")
+            task_aws_to_gcs_transfer = self.make_python_operator(self.aws_to_gcs_transfer, "aws_to_gcs_transfer")
+            task_branch = BranchPythonOperator(
```
keegansmith21, line 381: According to the internet, the BranchPythonOperator will push its chosen branch to Xcom by default each time it executes (and not delete the xcom afterwards). This can be changed by adding do_xcom_push=False. Not sure if we want the branch choice flooding the xcom, what do you think?

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -356,53 +363,70 @@ def __init__(
...
+                    >> task_check_dependencies
+                    >> task_fetch_releases
+                    >> task_create_datasets
+                    >> task_aws_to_gcs_transfer
+                    >> task_branch
+                    >> task_groups
+                    >> task_cleanup
+                    >> task_wait
+            )
+            # fmt: on
```
keegansmith21, line 427: Does black not like the airflow task linking?

jdddog: I found it tidier by having each task definition on a single line, although I could tighten up the part that isn't formatted.

academic_observatory_workflows/workflows/openalex_telescope.py
jdddog: The cleanup task will eventually remove the old XComs, so it should be OK, what do you think?

keegansmith21: Ah yes this is true, I forgot about the cleanup step :+1:

academic_observatory_workflows/workflows/openalex_telescope.py
```diff
@@ -597,180 +597,239 @@ def aws_to_gcs_transfer(self, release: OpenAlexRelease, **kwargs):
...
+
+        return task_ids
+
+    def bq_snapshot(self, release: OpenAlexRelease, entity_name: str = None, **kwargs):
+        """Create a snapshot of each main table. The purpose of this table is to be able to rollback the table
+        if something goes wrong. The snapshot expires after self.snapshot_expiry_days."""
+
+        task_id = get_task_id(**kwargs)
+        if release.is_first_run:
+            raise AirflowSkipException(make_first_run_message(task_id))
```
keegansmith21, line 625: I've tried these skipexections before, but in my testing they seem to skip everything downstream as well. Did you look into this?

jdddog: Yep. Tasks have a default trigger rule of `TriggerRule.SUCCESS`, which means that direct downstream tasks will only run if their direct upstream task was successful. To make the downstream task run on success or skipped, then the `TriggerRule.NONE_FAILED` trigger rule needs to be used.

keegansmith21: Ah I see! The other tasks must have skipped (in my testing) because that's the default behaviour. I can see that you've altered the downstream task trigger rule accordingly.

keegansmith21 (approved):
Thanks Jamie, these changes look really good!

PR #189, opened 2023-10-09
Fix ORCID telescope queue name

keegansmith21:
I put in the wrong name for the remote queue initially. This PR fixes it and also removes an unused variable in the class constructor.


jdddog (approved):
Thanks Keegan, looks good.

PR #190, opened 2023-10-11
python 3.10 update

keegansmith21:
Updates for python 3.10 compatibility. Depends on [observatory-platform update](https://github.com/The-Academic-Observatory/observatory-platform/pull/628)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [ 3.8 ]
+        python-version: [ 3.10 ]
         os: [ ubuntu-20.04 ]
 
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [ 3.8 ]
+        python-version: [ "3.10" ]
         os: [ ubuntu-20.04 ]
```
jdddog, line 12: Perhaps we should update these to use the latest version of Ubuntu:

```suggestion
        os: [ ubuntu-latest ]
```

jdddog (commented):
Thanks Keegan, the changes look good to me. I just wonder if we should change the OS that we use for tests, ubuntu-latest would be using Ubuntu-22.04.

jdddog (approved):
Thanks Keegan, looks good!

PR #195, opened 2023-10-23
Zenodo API changes

jdddog:
This is a fix for the ROR telescope, which broke because of breaking changes to the Zenodo API.


alexmassen-hane (approved):
Thanks Jamie, looks good.

keegansmith21 (approved):
Cheers Jamie. All looks good to me

PR #196, opened 2023-10-30
Use more SQL for OA Dashboard Workflow

jdddog:
This PR refactors the OA Dashboard workflow to use SQL rather than Pandas to produce the data for the dashboard, as this is a much easier and more efficient way to create the datasets and it uses less memory. Pandas was consuming too much memory.

The workflow now operates like this:
* Uploads a set of ROR IDs to BigQuery. These are IDs of institutions that are always included in the dashboard.
* Runs country.sql.jinja2 and institution.sql.jinja2, creating the country and institution tables in BigQuery.
* Downloads and post processes Wikipedia descriptions for countries (and then institutions). It updates the descriptions straight into the country and institution BigQuery tables. 
* Download's the institution logos and updates the paths for the logos straight into the country and institution BigQuery tables.
* Exports the country and institution tables to JSON lines format and downloads to disk.
* Does some basic post processing of the data to put it into the right form to be used by the website and uploaded to Zenodo, selecting only fields that are required for certain files (using glom). Adds some global stats and histogram calculations that are harder to do in SQL (perhaps that can be moved over to SQL later).
* The workflow continues the same as before from here.

I thought that putting the files associated with each workflow in the same place made it easier to see and find the SQL templates, schemas and tests used by the workflow. I created a module for the workflow, put the code for the workflow in this module and created folders for the SQL templates and BigQuery schemas. See below for an example. What do you think @alexmassen-hane and @keegansmith21?

image

Some of the helper functions such as `schema_folder` and `sql_folder` need updating, your suggestions would be welcome.

We would need to update the other workflows as well, although that could just be done over time.


alexmassen-hane (commented):
Thanks Jamie. I like the change to the directory structure for the workflow. Just minor nit comments on the PR.

Perhaps we could change the schema and sql functions to something like this? Or perhaps just do a check if the workflow folder exists and explicitly use that. 

```
def schema_folder(workflow_name: Optional[str] = None) -> str:
    """Return the path to the schema folder.

    :param workflow_name: Optional, name of the workflow. Only to be included if the schema for the workflow is in
        the directory academic_observatory_workflows.workflows.{workflow_name}.schema
    :return: the path.
    """

    if workflow_name:
        module_path = f"academic_observatory_workflows.workflows.{workflow_name}.schema"
        assert os.path.exists(
            module_file_path(module_path)
        ), f"Workflow name {workflow_name} given but schema folder within the workflow path does not exist!"
        return module_file_path(module_path)

    return module_file_path("academic_observatory_workflows.database.schema")
```

academic_observatory_workflows/workflows/oa_web_workflow/oa_web_workflow.py
```diff
@@ -0,0 +1,1045 @@
...
+            destination_uri = f"gs://{self.cloud_workspace.download_bucket}/{blob_prefix}/{entity_type}-data-*.jsonl.gz"
+            table_id = release.oa_dashboard_table_id(entity_type)
+            success = bq_query_to_gcs(
+                query=f"SELECT * FROM {table_id} ORDER BY stats.p_outputs_open DESC",  # Uses a query to export data to make sure it is in the correct order
+                project_id=self.output_project_id,
+                destination_uri=destination_uri,
+            )
+            results.append(success)
+
+        if not all(results):
```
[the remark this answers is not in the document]
jdddog: Apparently assertions can be turned off, so we shouldn't be using them for control flow, just exceptions. We might need to change some of our code. https://stackoverflow.com/a/21133264

jdddog:
> Thanks Jamie. I like the change to the directory structure for the workflow. Just minor nit comments on the PR.
> 
> Perhaps we could change the schema and sql functions to something like this? Or perhaps just do a check if the workflow folder exists and explicitly use that.
> 
> ```
> def schema_folder(workflow_name: Optional[str] = None) -> str:
>     """Return the path to the schema folder.
> 
>     :param workflow_name: Optional, name of the workflow. Only to be included if the schema for the workflow is in
>         the directory academic_observatory_workflows.workflows.{workflow_name}.schema
>     :return: the path.
>     """
> 
>     if workflow_name:
>         module_path = f"academic_observatory_workflows.workflows.{workflow_name}.schema"
>         assert os.path.exists(
>             module_file_path(module_path)
>         ), f"Workflow name {workflow_name} given but schema folder within the workflow path does not exist!"
>         return module_file_path(module_path)
> 
>     return module_file_path("academic_observatory_workflows.database.schema")
> ```

Thanks @alexmassen-hane, I've made those changes and updated the functions in config.py: https://github.com/The-Academic-Observatory/academic-observatory-workflows/blob/0e73b20780b1587ee62046b386c951b5432191e5/academic_observatory_workflows/config.py

Also added a function in this PR that is required: https://github.com/The-Academic-Observatory/observatory-platform/pull/647

academic_observatory_workflows/config.py
```diff
@@ -35,19 +36,59 @@ def test_fixtures_folder(*subdirs) -> str:
     return os.path.join(base_path, *subdirs)
 
 
-def schema_folder() -> str:
+def workflow_test_fixtures_path(workflow_module: str, *subdirs) -> str:
```
keegansmith21, line 39:
```suggestion
def workflow_test_fixtures_path(workflow_module: str, *subdirs: str) -> str:
```

academic_observatory_workflows/config.py
```diff
@@ -35,19 +36,59 @@ def test_fixtures_folder(*subdirs) -> str:
...
-    return module_file_path("academic_observatory_workflows.database.sql")
+    # New directory structure
+    if workflow_module is not None:
+        return construct_module_path("academic_observatory_workflows", "workflows", workflow_module, "sql")
+
+    # Old directory structure
+    return construct_module_path("academic_observatory_workflows", "database", "sql")
+
+
+def construct_module_path(*parts) -> str:
```
keegansmith21, line 86:
```suggestion
def construct_module_path(*parts: str) -> str:
```

academic_observatory_workflows/workflows/oa_dashboard_workflow/oa_dashboard_workflow.py
```diff
@@ -0,0 +1,1036 @@
+# Copyright 2021 Curtin University
```
keegansmith21, line 1:
```suggestion
# Copyright 2021-2023 Curtin University
```

academic_observatory_workflows/config.py
```diff
@@ -35,19 +36,59 @@ def test_fixtures_folder(*subdirs) -> str:
     return os.path.join(base_path, *subdirs)
 
 
-def schema_folder() -> str:
+def workflow_test_fixtures_path(workflow_module: str, *subdirs) -> str:
+    """Get the path to the Academic Observatory Workflows test data directory.
+
+    :param workflow_module: Optional, name of the workflow. Only to be included if the schema for the workflow is in
```
keegansmith21, line 42: As it stands, *workflow_module* is not optional. We could consolidate this function and *test_fixtures_folder()* in the same manner as the other functions such that it is optional. Unless we expect that the root level fixtures folder will soon become obsolete?

academic_observatory_workflows/workflows/oa_web_workflow/oa_web_workflow.py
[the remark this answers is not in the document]
keegansmith21: Yeah I looked into the use of assertions as well. Apparently they're only to be used in tests and elsewhere we should explicitly raise an error.

keegansmith21 (commented):
Awesome work updating this to use SQL. Also really like the suggested folder structure, it should compartmentalise stuff much nicer and make it easier to find the resources you need when working on a particular workflow.

Just a couple of very minor suggestions and a comment/question from me :+1:

academic_observatory_workflows/config.py
jdddog: Thanks Keegan, I've done this, `workflow_module` has to be a keyword argument after `*subdirs` if it is optional:

```python
def test_fixtures_folder(*subdirs: str, workflow_module: Optional[str] = None) -> str:
```

PR #198, opened 2023-11-10
INF-665: OpenAlex Telescope - Schema detect feature

alexmassen-hane:
As there are constantly issues with OpenAlex changing their data structure, I have added a feature that will scan through each transformed record to generate a Biguqery style schema. 

Tasks have been added that will upload the generated schema to the transform folder and compare it against the schema that we've defined for each entity. If there are any differences between the two it will send a Slack notification for the entities that have any differences.


jdddog (changes requested):
Hey Alex, thanks for the PR, it looks good, just a few small things to tweak below.

jdddog (approved):
Thanks Alex, the changes look good to me!

PR #200, opened 2023-11-15
Fix/oa dashboard

jdddog:
Fixes a couple of bugs introduced after update the OA Dashboard workflow.

The tests pass, it is the ORCID tests failing.


keegansmith21 (approved):
Thanks Jamie, changes look good!

PR #201, opened 2023-11-23
Feature/Update OpenAlex schema - November 2023 snapshot

alexmassen-hane:
OpenAlex has added some new fields to the works table and we have missed a few from funders. This PR adds the following to the appropriate schemas:

Funders:
- ids.crossref: string
- ids.doi: string

Works:
- keywords: repeated string
- cited_by_percentile_year.min: float
- cited_by_percentile_year.max: float

Unfortunately at this time the OpenAlex team has not yet added the new fields for the works table on their documentation page: https://docs.openalex.org/api-entities/works/work-object
I found the structure of the new fields by looking at records from their API.


jdddog (approved):
Thanks Alex, the changes look good.

PR #202, opened 2023-11-28
INF-638: Data Quality Workflow

alexmassen-hane:
# Data Quality Workflow

This PR introduces a data quality workflow for tables produced in the Academic Observatory.

## Metadata

This workflow will produce a table `project_id.data_quality.data_quality` containing metadata for singular table or a set of sharded tables, consisting of the following fields:

- full_table_id (str): Fully qualified id of the table in Google Bigquery f"{project_id}.{dataset_id}.{table_id}"
- hash_id (str): Based off of the full_table_id, number of fields, number of rows and number of bytes of stored in the table.
- project_id (str): Name of the Google project.
- dataset_id (str): Name of the dataset.
- table_id (str): Friendly table name.
- sharded (bool): If the table is sharded or not. 
- date_shard (date): Date from the table shard (null if it is not a sharded table).
- date_created (timestamp): Date of when the table was created. 
- expires (bool): If the table is set to expire or not. 
- date_expires (timestamp): Date of when the table expires, null if it not to expire.
- date_last_modified (timestamp): Date of when the table was modified.
- date_checked (timestamp): Date of when the table was checked by this workflow.
- size_gb (float): The size of the table in Bigquery in gigabytes. 
- primary_key ( List[str] ): The fields/keys used to identify records of a table. 
- num_rows (int): Number of rows in the table.
- num_distinct_records (int): Number of records with distinct primary key/s.
- num_null_records (int): Number of records that have no value for primary_key/s.
- num_duplicates (int): Number of records that have duplicate values under the primary_key/s.
- num_all_fields (int): Total number of fields in the table. This is see how the table grows/modified overtime. 

Notes:
- The last few fields of the metadata records; num_distinct_records, num_null_records and num_duplicates are determined by using small SQL quires that run using the Google API. 
- This workflow was designed to be as lightweight as possible. It does not require a connection to the Observatory API and it uses the hash_id to determine if it needs to perform the check a on table again.
- The hash_id is created using a combination of the name of the table, number of bytes stored, number of rows and the total number of fields. This is a cheap and quick way of roughly determining a similar md5hash. However, this assumes that changes to the data in the table is a significant number of bytes and rows different than it's previous version. Upsert and delete operations typically change tables on the order of megabytes and 1000s of rows so the chances of the table hash_ids being the same is unlikely. 
- This workflow includes the ability to check over datasets that are not just produced by AO Workflows, for example OpenAIRE, which has been manually added to the project and will be checked just the same as other tables.

## Workflow details

This workflow is designed so that there will be an instance for each dataset and the checks for the subsequent tables are run in parallel, for example:

Untitled

Each workflow instance is defined by the following in the config.yaml for the Observatory Platform. 

Notes:
-  The "shard_limit" param can be set to False if you want all shards analysed by the workflow, or set to an integer "x" which will only check over the most "x" recent shards. If it is not included in the config, it will default to the value of 5.
- For manually uploaded datasets such as OpenAIRE, simply remove the schedule param form the config to remove the dag_sensor for the workflow.

Example config for a dataset that has regularly updated tables:

```
- dag_id: "data_quality_openalex"
    name: "Data Quality - OpenAlex"
    class_name: "academic_observatory_workflows.workflows.data_quality_workflow.DataQualityWorkflow"
    cloud_workspace: *academic_observatory
    kwargs:
      schedule: "@weekly"
      sensor_dag_ids: ["openalex"]
      datasets:
        openalex:
          tables:
            - table_id: authors
              primary_key: [id]
              is_sharded: false
            - table_id: works
              primary_key: [id]
              is_sharded: false
```

Example config for a dataset with a collection of table shards:

```
- dag_id: "data_quality_ror"
    name: "Data Quality - ROR"
    class_name: "academic_observatory_workflows.workflows.data_quality_workflow.DataQualityWorkflow"
    cloud_workspace: *academic_observatory
    kwargs:
      schedule: "@weekly"
      sensor_dag_ids: ["pubmed"] 
      datasets:
        pubmed:
          tables:
            - table_id: authors
              primary_key: [MedlineCitation.PMID.value, MedlineCitation.PMID.Version]
              is_sharded: True
              shard_limit: False
```

### DIFF
--- a/academic_observatory_workflows/config.py
+++ b/academic_observatory_workflows/config.py
@@ -24,6 +24,7 @@ class Tag:
     """DAG tag."""
 
     academic_observatory = "academic-observatory"
+    data_quality = "data-quality"
 
 
 def test_fixtures_folder(*subdirs: str, workflow_module: Optional[str] = None) -> str:
--- a/academic_observatory_workflows/database/schema/data_quality/data_quality.json
+++ b/academic_observatory_workflows/database/schema/data_quality/data_quality.json
@@ -1,0 +1,116 @@
+[
+  {
+    "name": "full_table_id",
+    "description": "The fully qualified table name: project_id.dataset_id.table_id",
+    "type": "STRING",
+    "mode": "REQUIRED"
+  },
+  {
+    "name": "hash_id",
+    "description": "A unique table identifier based off of the full_table_id, number of bytes, number of rows and numbre of columms.",
+    "type": "STRING",
+    "mode": "REQUIRED"
+  },
+  {
+    "name": "project_id",
+    "description": "Name of the project.",
+    "type": "STRING",
+    "mode": "REQUIRED"
+  },
+  {
+    "name": "dataset_id",
+    "description": "Dataset that holds the table.",
+    "type": "STRING",
+    "mode": "REQUIRED"
+  },
+  {
+    "name": "table_id",
+    "description": "Name of the table.",
+    "type": "STRING",
+    "mode": "REQUIRED"
+  },
+  {
+    "name": "shard_date",
+    "description": "Date from the table shard (if null it is not a sharded table).",
+    "type": "DATE",
+    "mode": "NULLABLE"
+  },
+  {
+    "name": "is_sharded",
+    "description": "If the table is sharded or not.",
+    "type": "BOOL",
+    "mode": "REQUIRED"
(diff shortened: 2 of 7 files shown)


jdddog (approved):
Thanks for the PR Alex, it looks good to me.

PR #203, opened 2023-11-29
Fix: OpenAlex Schema Generator

alexmassen-hane:
This PR is a proposed fix for the previously introduced schema generator feature. Unfortunately, previous version for merging schemas continually added to the grand merged schema (a nested dictionary object) for each part data file transformed by the workflow and caused the worker VM to unexpectedly crash.

Instead of using the 'mergedeep' library to merge each schema found from each part data file, it uses the schema_map generated for each file and merges them together using the 'merge_schema_entry' on the SchemaGenerator class. This is flattened to a human readable format and compared against our pre-defined schemas for each entity. I have also added a test to ensure that the merging of the schemas is done properly and does not miss or replace any fields. 

Please review and edit as necessary.


jdddog (approved):
Thanks for fixing this Alex!

PR #204, opened 2023-12-01
Update: Fix OpenAlex schema and add missing fields

alexmassen-hane:
This PR is to fix the works table schema so that it can be ingested into BigQuery successfully. While I was debugging the failed import, I found a number of fields that we were missing in our schema. Most of the missing fields were under the locations and sources dehydrated objects and I have included them into this PR. 

Please review and edit as necessary.


jdddog (approved):
Thanks for updating the schema Alex, it looks good.

PR #205, opened 2023-12-13
Optimise OpenAlex table loading

jdddog:
Optimised OpenAlex table loading and retention in several ways:
* Load data straight into the main table on the first run, rather than into the upsert table and then the main table.
* Make the upsert and delete tables expire after a period of 7 days (default value).

Also:
* Updated tests.
* Changed asserts to raising Exceptions, as asserts should not be used for control flow and can be turned off.


alexmassen-hane (approved):
Thanks Jamie, looks good.

PR #207, opened 2023-12-18
Fix/PubMed: No updatefiles found in release period

alexmassen-hane:
Problem:
- The workflow had it's release_end_date set to the execution_date, meaning that it was set to a week before it should have been for the release. This caused an issue when looking for updatefiles after the latest baseline dump as the release_start_date (the baseline upload date, 2023-12-14) was set to be after release_interval_end (2023-12-10)  which meant that no updatefiles were found , causing the workflow to fail. The dates for the release period should have been between 2023-12-14 and 2023-12-17 (the actual run date that the workflow was run).

Fix: 
- Change the release_interval_end variable to use data_interval_end,  and added a check to make sure that the baseline_upload_date is before the release_interval_end, otherwise raise an AirflowException. 
- Added skips for the edge case when the workflow runs on the day that the baseline annual release is uploaded and  there are no updatefiles available for the release period.


jdddog (approved):
Thanks for fixing the workflow Alex!

PR #209, opened 2024-01-10
Fix/OpenAlex: Account for new top level field when merging generated schemas

alexmassen-hane:
The merge schema function in the OpenAlex telescope only accounted for sub/nested fields in incoming the data structure. This fix accounts for new top level fields that are not present in the "old" schema and simply adds them to the merged schema object. I have also added this specific case to the unit test files. 

Please review and edit as needed.


keegansmith21 (approved):
Thanks for doing this Alex!

PR #213, opened 2024-01-25
Increase OpenAlex upsert byte limit

jdddog:
Increase OpenAlex upsert byte limit


keegansmith21 (commented):
Thanks Jamie. Just wondering why use 2^40 for the bytes limit?

keegansmith21 (approved):

PR #215, opened 2024-03-05
Restructure workflow layout

jdddog:
Restructured the workflows into their own modules, with sql and schema folders inside.

Deleted Crossref Events and Geonames as these are no longer used.


keegansmith21 (approved):
Thanks Jamie. I really like the project_path function you've introduced here!

PR #217, opened 2024-03-14
Fix fetch wikipedia descriptions

jdddog:
Fixed an issue in `fetch_wikipedia_descriptions_batch`, which was not returning descriptions for all input URLs in all cases. Specifically, when duplicate Wikipedia URLs were supplied and when two different URLs that point to the same page were supplied (e.g. one URL encoded and one not) .

Also added logging messages to print more info when there is a discrepancy between input URLs and output URLs.


keegansmith21 (approved):
Thanks Jamie, changes look good!

PR #223, opened 2024-06-17
Fix/workflow bugs

jdddog:
This PR fixes bugs in a few workflows.

Unpaywall:
* Unpaywall changefile changed_dois_with_versions_2024-06-10T080001.jsonl at row 82563 is just null instead of a valid JSON object.
* Changfile sorting was in the release object was incorrect.

OA Dashboard workflow:
* add_wiki_descriptions tasks for country and institution were not running properly.

OpenAlex:
* Tasks were not skipping when an entity was not updated.
* The works table requires a higher query limit.

DOI workflow:
* Comment out regions aggregation as it has been getting a `Resources exceeded during query execution: The query could not be executed in the allotted memory. Peak usage: 100% of limit.` error for some time.


academic_observatory_workflows/openalex_telescope/openalex_telescope.py
```diff
@@ -69,7 +69,7 @@
 )
 
 TEMP_TABLE_DESCRIPTION = "Temporary table for internal use. Do not use."
-UPSERT_BYTE_LIMIT = int(2.5 * 2**40)
+UPSERT_BYTE_LIMIT = int(3 * 2**40)
```
keegansmith21, line 72: I forget, does AO have a config file? If so maybe we could supply the byte limit through it as it seems to be changing fairly frequently?

academic_observatory_workflows/unpaywall_telescope/unpaywall_telescope.py
```diff
@@ -608,12 +608,14 @@ def transform(release: dict, **context):
                     "transform: find and replace the 'authenticated-orcid' string in the jsonl to 'authenticated_orcid'"
                 )
                 for changefile in release.changefiles:
-                    find_replace_file(
-                        src=changefile.extract_file_path,
-                        dst=changefile.transform_file_path,
-                        pattern="authenticated-orcid",
-                        replacement="authenticated_orcid",
-                    )
+                    with open(changefile.extract_file_path, "r") as file_in:
```
keegansmith21, line 611: This might look nicer with both open() statements in the same context manger

```
with open(file1, "r") as f_in, open(file2, "w") as f_out:
```
Up to you though

keegansmith21 (approved):
Thanks for these changes Jamie. 

Just a couple of comments but feel free to merge.

academic_observatory_workflows/openalex_telescope/openalex_telescope.py
jdddog: True, if I make it as an argument for the workflow with a default value, then we can specify it in the config file.

academic_observatory_workflows/unpaywall_telescope/unpaywall_telescope.py
jdddog: Thanks, I'll change that

PR #224, opened 2024-07-09
Feature/astro2

keegansmith21:


academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+        self.upload_downloaded = V1ResourceRequirements(
+            requests=task_resources["upload_downloaded"], limits=task_resources["upload_downloaded"]
+        )
+        self.extract = V1ResourceRequirements(requests=task_resources["extract"], limits=task_resources["extract"])
+        self.transform = V1ResourceRequirements(
+            requests=task_resources["transform"], limits=task_resources["transform"]
+        )
+        self.upload_transformed = V1ResourceRequirements(
+            requests=task_resources["upload_transformed"], limits=task_resources["upload_transformed"]
+        )
```
jdddog, line 65: I think it would be better to make something that can be used in a more generic way and for each kubernetes task decorator, e.g.

```python
def make_container_resources(default: dict, override: Optional[dict]) -> V1ResourceRequirements:
    resource = default
    if override is not None:
        resource = override
    return V1ResourceRequirements(requests=resource, limits=resource)
```

```python
@task.kubernetes(
    name="extract",
    container_resources=make_container_resources(
        {"memory": "8Gi", "cpu": "8"}, dag_params.gke_resource_overrides.get("extract")
    ),
    **kubernetes_task_params,
)
def extract(release: dict, **context):
    """Extracts the compressed data"""

    from academic_observatory_workflows.crossref_metadata_telescope.tasks import extract

    extract(release, **context)
```

You just have to change the `gke_resource_overrides` variable to: `gke_resource_overrides: dict = field(default_factory=dict)`

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+    :param gke_image: The image location to pull from.
+    :param gke_namespace: The cluster namespace to use.
+    :param gke_startup_timeout_seconds: How long to wait for the container to start in seconds.
+    :param gke_conn_id: The name of the airlfow connection storing the gke cluster information.
+    :param docker_astro_uuid: The uuid of the astro user
+    :param gke_volume_name: The name of the persistent volume to create
+    :param gke_volume_path: Where to mount the persistent volume
+    :param gke_zone: The zone containing the gke cluster
+    :param gke_volume_size: The amount of storage to request for the persistent volume in GiB
+    :param gke_resource_overrides: Task resource overrides
```
jdddog, line 99: Could these be in their own data class and added as a single parameter to this one? Since they will be re-used across a number of different workflows, with the same settings.

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+        get_logs=True,
+        in_cluster=False,
+        kubernetes_conn_id=dag_params.gke_conn_id,
+        log_events_on_failure=True,
+        namespace=dag_params.gke_namespace,
+        startup_timeout_seconds=dag_params.gke_startup_timeout_seconds,
+        env_vars={"DATA_PATH": dag_params.gke_volume_path},
+        volumes=volumes,
+        volume_mounts=volume_mounts,
+    )
```
jdddog, line 165: I would make this into a reusable function, since it will be used for each workflow. This is also why putting the GKE related params into their own data class might be good, then `dag_params: DagParams` could be changed to that data class specifically.  

```python
def make_kubernetes_task_params(dag_params: DagParams):
    volume_mounts = [k8s.V1VolumeMount(mount_path=dag_params.gke_volume_path, name=dag_params.gke_volume_name)]
    volumes = [
        k8s.V1Volume(
            name=dag_params.gke_volume_name,
            persistent_volume_claim=k8s.V1PersistentVolumeClaimVolumeSource(claim_name=dag_params.gke_volume_name),
        )
    ]
    return dict(
        image=dag_params.gke_image,
        security_context=k8s.V1PodSecurityContext(
            fs_group=dag_params.docker_astro_uid,
            fs_group_change_policy="OnRootMismatch",
            run_as_group=dag_params.docker_astro_uid,
            run_as_user=dag_params.docker_astro_uid,
        ),
        do_xcom_push=True,
        get_logs=True,
        in_cluster=False,
        kubernetes_conn_id=dag_params.gke_conn_id,
        log_events_on_failure=True,
        namespace=dag_params.gke_namespace,
        startup_timeout_seconds=dag_params.gke_startup_timeout_seconds,
        env_vars={"DATA_PATH": dag_params.gke_volume_path},
        volumes=volumes,
        volume_mounts=volume_mounts,
    )
```

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+            from academic_observatory_workflows.crossref_metadata_telescope.tasks import fetch_release
+
+            return fetch_release(
+                run_id=context["run_id"],
+                data_interval_start=context["data_interval_start"],
+                data_interval_end=context["data_interval_end"],
+                cloud_workspace=dag_params.cloud_workspace,
+                crossref_metadata_conn_id=dag_params.crossref_metadata_conn_id,
+                dag_id=dag_params.dag_id,
+            )
```
jdddog, line 194: Do these need to start with an underscore? It is cleaner without them.

I think the Kubernetes based task functions and functions within them can have the same name? 

For tasks like these that run as Python operators, so that there is no conflict, and so that there are less imports, you could instead import the tasks module once globally, alias it and then reuse it:

```python
from academic_observatory_workflows.crossref_metadata_telescope.tasks as tasks

@task
def fetch_release(dag_params: DagParams, **context) -> dict:
    """Fetch the release for this month, making sure that it exists."""

    return tasks.fetch_release(
        run_id=context["run_id"],
        data_interval_start=context["data_interval_start"],
        data_interval_end=context["data_interval_end"],
        cloud_workspace=dag_params.cloud_workspace,
        crossref_metadata_conn_id=dag_params.crossref_metadata_conn_id,
        dag_id=dag_params.dag_id,
    )
```

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+        @task.kubernetes(
+            name="upload_download",
+            container_resources=task_resources.upload_downloaded,
+            **kubernetes_task_params,
+        )
+        def _upload_downloaded(release: dict, dag_params, **context):
+            """Upload downloaded data to Cloud Storage."""
+            from academic_observatory_workflows.crossref_metadata_telescope.tasks import upload_downloaded
+
+            upload_downloaded(release, cloud_workspace=dag_params.cloud_workspace)
```
jdddog, line 217: `cloud_workspace` parameter doesn't exist

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+        @task.kubernetes(
+            name="upload_transformed",
+            container_resources=task_resources.upload_transformed,
+            **kubernetes_task_params,
+        )
+        def _upload_transformed(release: dict, dag_params, **context) -> None:
+            """Uploads the transformed data to cloud storage"""
+            from academic_observatory_workflows.crossref_metadata_telescope.tasks import upload_transformed
+
+            upload_transformed(release, cloud_workspace=dag_params.cloud_workspace)
```
jdddog, line 251: `cloud_workspace` parameter doesn't exist

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+
+        @task
+        def _bq_load(release: dict, dag_params: DagParams, **context):
+            """Loads the data into a bigquery table"""
+            from academic_observatory_workflows.crossref_metadata_telescope.tasks import bq_load
+
+            bq_load(
+                release,
+                cloud_workspace=dag_params.cloud_workspace,
```
jdddog, line 260: `cloud_workspace` parameter doesn't exist

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,326 @@
...
+
+        @task
+        def _add_dataset_release(release: dict, dag_params: DagParams, **context) -> None:
+            """Adds a release to the dataset API"""
+            from academic_observatory_workflows.crossref_metadata_telescope.tasks import add_dataset_release
+
+            add_dataset_release(
+                release,
+                dag_id=dag_params.dag_id,
+                cloud_workspace=dag_params.cloud_workspace,
```
jdddog, line 276: `dag_id` and `cloud_workspace` parameters don't exist

jdddog (changes requested):
Hey Keegan, the changes are looking really good, it is much tidier and easier to read the Airflow workflow now that the tasks are split into their own file. I've left a few comments about things that could potentially be refined.

deploy.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
```
jdddog, line 2: It would be good to move all of the files that we made related to deployment, running test environments etc into `bin` to keep the root directory tidy.

academic-observatory-workflows/academic_observatory_workflows/ror_telescope/tests/test_tasks.py
```diff
@@ -0,0 +1,102 @@
...
+import pendulum
+import vcr
+
+from academic_observatory_workflows.config import project_path
+from academic_observatory_workflows.ror_telescope.tasks import list_ror_records, is_lat_lng_valid, transform_ror
+
+FIXTURES_FOLDER = project_path("ror_telescope", "tests", "fixtures")
+
+
+class TestListRorRecords(unittest.TestCase):
```
jdddog, line 13: Group the tests into same class.

academic-observatory-workflows/academic_observatory_workflows/pubmed_telescope/telescope.py
```diff
@@ -0,0 +1,539 @@
...
+        bq_delete_table_name: str = "pubmed_delete",
+        bq_dataset_description: str = "Pubmed Medline database, only PubmedArticle records: https://pubmed.ncbi.nlm.nih.gov/about/",
+        start_date: pendulum.DateTime = pendulum.datetime(year=2021, month=1, day=1),
+        schedule: str = "@weekly",
+        ftp_server_url: str = "ftp.ncbi.nlm.nih.gov",
+        ftp_port: int = 21,
+        reset_ftp_counter: int = 40,
+        max_download_attempt: int = 5,
+        snapshot_expiry_days: int = 31,
+        max_processes: int = 4,  # Limited to 4 due to RAM usage.
```
jdddog, line 76: May not be necessary anymore now that we are using Kubernetes and the transform task will have all of the RAM to itself. If not enough then we can increase the RAM and keep the CPUs the same.

I would then remove this parameter and us os.cpu_count() in this function in tasks.py:

```python
def updatefiles_merge_upserts_deletes(release: dict, updatefiles: List[Dict], max_processes: int = os.cpu_count()) -> None:
```

We don't want to use os.cpu_count() in the DAG, as it will then set the CPU count based on the Astro pod where the DAG is loaded I think.

We also need to remove / fix this same thing with Crossref Metadata, as it is setting os.cpu_count in the DagParams too.

academic-observatory-workflows/academic_observatory_workflows/orcid_telescope/telescope.py
```diff
@@ -0,0 +1,355 @@
...
+        transfer_attempts: int = 5,
+        max_workers: Optional[int] = None,
+        observatory_api_conn_id: str = AirflowConns.OBSERVATORY_API,
+        aws_orcid_conn_id: str = "aws_orcid",
+        start_date: pendulum.DateTime = pendulum.datetime(2023, 6, 1),
+        schedule: str = "0 0 * * 0",  # Midnight UTC every Sunday
+        max_active_runs: int = 1,
+        retries: int = 3,
+        test_run: bool = False,
+        gke_volume_size: int = 1000,
```
jdddog, line 92: gke_volume_size should be changed to a string with units.

jdddog (commented):
Hey Keegan, thanks for the PR, it is looking really good, it was a lot of work to port these to use Kuberentes.

I've just got some small suggestions, I might have missed applying some of the comments for specific workflows to all of them, so it would be good to check if the other workflows / tests could also benefit from those changes.

There are a few comments for me to address too as I've noticed a few of things.

academic-observatory-workflows/academic_observatory_workflows/crossref_metadata_telescope/telescope.py
```diff
@@ -0,0 +1,285 @@
...
+        def extract(release: dict, **context):
+            """Extracts the compressed data"""
+            from academic_observatory_workflows.crossref_metadata_telescope import tasks
+
+            tasks.extract(release, **context)
+
+        @task.kubernetes(
+            name="transform",
+            container_resources=gke_make_container_resources(
+                {"memory": "2G", "cpu": "2"}, dag_params.gke_params.gke_resource_overrides.get("transform")
```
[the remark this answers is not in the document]
jdddog: The os.cpu_count issue also needs fixing here.

academic-observatory-workflows/academic_observatory_workflows/crossref_fundref_telescope/tests/test_tasks.py
```diff
@@ -0,0 +1,280 @@
...
+from observatory_platform.google.gcs import gcs_blob_name_from_path, gcs_upload_files
+from observatory_platform.sandbox.sandbox_environment import SandboxEnvironment
+from observatory_platform.sandbox.test_utils import SandboxTestCase
+from observatory_platform.files import load_jsonl, get_file_hash
+
+FIXTURES_FOLDER = project_path("crossref_fundref_telescope", "tests", "fixtures")
+SCHEMA_FOLDER = project_path("crossref_fundref_telescope", "schema")
+
+
+class TestListReleases(unittest.TestCase):
```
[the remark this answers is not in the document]
keegansmith21: Yeah I went back and forth on this. Here's my thinking:
- When the tasks are simple and don't require complicated, numerous tests, then It's nicer to group them under the same class because they have no reason to be separate
- When the tasks are not simple (require multiple test functions for branching logic coverage and/or benefit from class level variables), then grouping the different tasks under the same class decreases readability as unrelated variables are declared simultaneously and can be accessed by the functions far from where they're declared.
- Since some tests are better with grouping, and some are better without grouping, this breaks consistency. So to maintain it I just made all tasks ungrouped, which makes some tests unfortunately verbose.

What are your thoughts on this?

academic-observatory-workflows/academic_observatory_workflows/pubmed_telescope/telescope.py
keegansmith21: Oh good catch! Yeah i believe cpu_count() will be done at dag creation time on the astro machine. Should be fixable by simply putting it in the task itself which will be done at runtime.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN apt-get update && apt-get install -y google-cloud-cli google-cloud-cli-gke-gcloud-auth-plugin
+
+# ORCID install s5cmd
+RUN curl -LO https://github.com/peak/s5cmd/releases/download/v2.1.0/s5cmd_2.1.0_linux_amd64.deb && dpkg -i s5cmd_2.1.0_linux_amd64.deb
+
+USER astro
+
+# Install Observatory Platform
+# RUN git clone --branch feature/astro-refactor https://github.com/The-Academic-Observatory/observatory-platform.git
+RUN git clone --branch feature/astro_kubernetes https://github.com/The-Academic-Observatory/observatory-platform.git
```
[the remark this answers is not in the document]
keegansmith21: Yep I've changed it in the unpaywall PR. Should make a note to switch it to main when it becomes available

deploy.sh
keegansmith21: Sure good idea. I moved most of the test/deploy files into bin but the 3 Dockerfiles need to stay in root as docker context can't move to parent directories.

academic-observatory-workflows/academic_observatory_workflows/pubmed_telescope/telescope.py
jdddog: Yeah that works, thanks for fixing it.

academic-observatory-workflows/academic_observatory_workflows/crossref_fundref_telescope/tests/test_tasks.py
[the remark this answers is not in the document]
jdddog: I would group the simple tasks into the same TestTasks class, and put more complicated tasks into their own test classes. Otherwise it creates more boilerplate code that you have to create and maintain.

PR #225, opened 2024-07-19
Skip cases where a relationship org is a both parent and a child

jdddog:


keegansmith21 (approved):
Looks good, Thanks Jamie!

PR #226, opened 2024-07-28
Fix OA Dashboard divide by zero null error

jdddog:
Return 0 for percentage rather than NULL when divide by zero error occurs.


keegansmith21 (approved):
Good pickup, thanks Jamie!

PR #227, opened 2024-08-15
Remove execution_delta from PreviousDagRunSensor

jdddog:


keegansmith21 (commented):
Thanks Jamie. Are you planning on adding the newly created execution delta function to these in a future update?

jdddog:
> Thanks Jamie. Are you planning on adding the newly created execution delta function to these in a future update?

Thanks Keegan. It is just hard coded as a part of the PreviousDagRunSensor, so nothing needs to be passed to it.

keegansmith21 (approved):
Looks good, Thanks Jamie!

PR #228, opened 2024-08-22
Add DataCite schema and transformation script

jdddog:
Scripts, JSON schema and documentation for manually loading the [DataCite Public Data File](https://datafiles.datacite.org/datafiles/public-2023) into BigQuery.


keegansmith21 (approved):
Thanks Jamie!

PR #232, opened 2024-11-04
Re-order new affiliations field in works table schema

jdddog:
Re-order new affiliations field in works table schema so that fields in works upsert table have same order as works table


keegansmith21 (approved):
Thanks Jamie!

PR #234, opened 2024-11-15
Feature/astro2 jamie

jdddog:
Refactored the the DOI Workflow and OA Dashboard Workflow to fo

Refactored the OpenAlex Telescope to with the Kubernetes task decorator.


academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/release.py
```diff
@@ -0,0 +1,251 @@
...
+            bq_dataset_id=self.bq_dataset_id,
+            schema_folder=self.schema_folder,
+            snapshot_date=self.snapshot_date.isoformat(),
+            manifest=self.manifest.to_dict(),
+            merged_ids=[merged_id.to_dict() for merged_id in self.merged_ids],
+            is_first_run=self.is_first_run,
+        )
+
+
+class Meta:
```
keegansmith21, line 123: Might be good to have a short docstring explaining what these classes represent

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/tasks.py
```diff
@@ -0,0 +1,683 @@
...
+    # Read in the expected schema for the entity.
+    merged_schema = load_json(entity.generated_schema_path)
+    expected_schema = load_json(entity.schema_file_path)
+
+    try:
+        match = bq_compare_schemas(expected_schema, merged_schema, check_types_match=False)
+    except:
+        match = False
+
+    if not match:
```
keegansmith21, line 273: Will this eventually lead to an error?

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/tasks.py
```diff
@@ -0,0 +1,683 @@
...
+    if not success:
+        raise AirflowException("upload_files: error uploading files to cloud storage")
+
+
+def bq_load_table(*, entity: OpenAlexEntity):
+    table_name = "main"
+    table_id = entity.bq_table_id
+    description = entity.table_description
+
+    if bq.bq_table_exists(table_id):
```
keegansmith21, line 299: Could you set the write disposition to WRITE_EMPTY? Then you wouldn't need to put this check in.

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/telescope.py
```diff
@@ -0,0 +1,502 @@
...
+        dataset_description: str = "The OpenAlex dataset: https://docs.openalex.org/",
+        non_concurrent_table_expiry_days: int = 62,
+        n_transfer_trys: int = 3,
+        primary_key: str = "id",
+        aws_conn_id: str = "aws_openalex",
+        aws_openalex_bucket: str = "openalex",
+        slack_conn_id: Optional[str] = AirflowConns.SLACK,
+        start_date: pendulum.DateTime = pendulum.datetime(2021, 12, 1),
+        schedule: str = "@weekly",
+        queue: str = "remote_queue",
```
keegansmith21, line 58: Is the remote queue still relevant? I think we can ditch it

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/telescope.py
```diff
@@ -0,0 +1,502 @@
...
+            entity_index: dict,
+            entity_name: str,
+            dag_params: DagParams,
+            gke_params: GkeParams,
+        ):
+            @task()
+            def aws_to_gcs_transfer(entity_index: dict, entity_name: str, dag_params: DagParams, **context):
+                """Transfer files from AWS bucket to Google Cloud bucket"""
+
+                import academic_observatory_workflows.openalex_telescope.tasks as tasks
```
keegansmith21, line 275: you can just import the tasks globally and call them as you normally would for the non-k8s tasks

keegansmith21 (commented):
Thanks Jamie, just a few comments

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/telescope.py
jdddog: Yeah you're right, I've removed it.

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/tasks.py
jdddog: That's a good point, I've changed that.

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/release.py
jdddog: Yeah that's fair enough, I've added docstrings to these classes.

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/tasks.py
jdddog: It is just for informational purposes, it compares the current schema to the generated schema based on the data that it loaded.

If new fields are added the data will still load, as we ignore new fields.

If there is a breaking change and the table doesn't load, then we can use the generated schema to help figure out why.

academic-observatory-workflows/academic_observatory_workflows/openalex_telescope/telescope.py
jdddog: I was thinking about this as I wrote the workflow, and I think that you were right for doing this originally for workflows with mixed kubernetes and normal tasks, because otherwise you don't get a warning that you haven't imported `tasks` into the kubernetes decorated function.

keegansmith21: Ah that's a good point!

PR #237, opened 2025-02-20
CI testing environment

keegansmith21:
Fixes some small issues in tests. More importantly, adds the minikube test environment to the CI tests.

This was very painful, exaggerated by inconsistencies in the minikube hostname behaviour: https://github.com/kubernetes/minikube/issues/8439 and the fact that the act runner does not fully isolate from its host environment... Which led to very long development time and some chown fixes. Hopefully nobody has to touch this again because it's a house of cards.


keegansmith21:
Hi @jdddog This is now ready for review. Apologies for the mess of commits - this was quite a fiddly task.

requirements.txt
```diff
@@ -1,2 +1 @@
 # Astro Runtime includes the following pre-installed providers packages: https://docs.astronomer.io/astro/runtime-image-architecture#provider-packages
-apache-airflow-providers-cncf-kubernetes>=8.3.2
```
jdddog, line 2: Do we not need this because it is installed by default when running Airflow from the Astro Docker container?

.github/workflows/unit-tests.yml
```diff
@@ -1,51 +1,84 @@
...
           cd academic-observatory-workflows
-          pip install -e .[tests] --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.7.3/constraints-no-providers-${{ matrix.python-version }}.txt
+          pip install -e ./academic-observatory-workflows[test] --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.7.3/constraints-no-providers-${{ matrix.python-version }}.txt
 
           cd ..
           curl -LO https://github.com/peak/s5cmd/releases/download/v2.1.0/s5cmd_2.1.0_linux_amd64.deb
           sudo dpkg -i s5cmd_2.1.0_linux_amd64.deb
 
+          pip freeze
+          airflow db init
```
jdddog, line 75: Why do we need to init the Airflow DB? Or is this because we are using the Airflow's dag.test?

jdddog (commented):
Thanks Keegan, it looks good to me, just a couple of questions / comments.

requirements.txt
keegansmith21: No it's because we already have the kubernetes package as part of observatory-platform. Having it here too seemed to result in some conflicts

.github/workflows/unit-tests.yml
keegansmith21: Yeah this was an issue I was having. I couldn't interact with the database at all and it didn't seem to exist so I added this line and it was happy. No idea why it isn't necessary when running locally but is required here.

jdddog (approved):

PR #239, opened 2025-03-27
Upload task merge

keegansmith21:
Combine the upload tasks into other tasks and download the file before task execution when needed.


jdddog:
Thanks Keegan, we only needed to combine the upload task for Crossref Fundref as it runs on Astro, as we can't guarantee that the same storage is available for the next task.

For the workflows with tasks that run on GKE, it is better that their download, extract and transform tasks remain separate, so that if there is a failure, only the individual task that fails need to be retried. The uploaded downloaded tasks can also be removed altogether.

